PR #17, opened 2017-09-07
OpenAPI 3 support

tmaxwell-stripe:
Consume an OpenAPI 3 spec instead of an OpenAPI 2 spec.


tmaxwell-stripe:
This passes `make test`, but I haven't tested it against a real client. I'm planning to check out https://github.com/stripe/stripe-go/pull/446 and see if that works.

brandur-stripe:
Whoah sweet. Will take a look in a bit!

> This passes make test, but I haven't tested it against a real client. I'm planning to check out stripe/stripe-go#446 and see if that works.

That should work, or just `master` in stripe-ruby.

spec/spec.go
```diff
@@ -2,81 +2,111 @@ package spec
...
+  type schema2 Schema
+  var inner schema2
+  err = json.Unmarshal(data, &inner)
+  if err != nil {
+    return err
+  }
+  *s = Schema(inner)
 
-	return nil
+  return nil
```
brandur-stripe: Possibly some indentation problems on this line and above.

spec/spec_test.go
```diff
@@ -1 +1,47 @@
...
+
+  assert "github.com/stretchr/testify/require"
+)
+
+func TestUnmarshal_Simple(t *testing.T) {
+  data := []byte(`{"type": "string"}`)
+  var schema Schema
+  err := json.Unmarshal(data, &schema)
+  assert.NoError(t, err)
+  assert.Equal(t, "string", schema.Type)
```
brandur-stripe: And same in this file and `validate.go`. FYI, `gofmt -w <file>` if you didn't know about it already.

generator.go
```diff
@@ -250,26 +242,20 @@ func duplicateMap(dataMap map[string]interface{}) map[string]interface{} {
...
-			pointer)
-	}
-
-	if len(parts) > 3 {
-		return "", fmt.Errorf("Pointer too long to be handle %v", pointer)
+	if len(parts) != 4 ||
+		parts[0] != "#" ||
+		parts[1] != "components" ||
+		parts[2] != "schemas" {
+		return "", fmt.Errorf("Expected '#/components/schemas/...' but got '%v'", pointer)
```
brandur-stripe: Nice.

param/coercer/coercer.go
```diff
@@ -29,21 +29,21 @@ func CoerceParams(schema *spec.JSONSchema, data map[string]interface{}) {
 		valStr, ok := val.(string)
 		if ok {
 			switch {
-			case hasType(subSchema, booleanType):
+			case subSchema.Type == booleanType:
```
brandur-stripe: Good simplification.

server.go
```diff
@@ -189,24 +189,30 @@ func (s *StubServer) initializeRouter() error {
...
-				return err
+			requestBodySchema := getRequestBodySchema(operation)
+			var requestBodyValidator *jsval.JSVal
+			if requestBodySchema != nil {
+				var err error
+				requestBodyValidator, err =
+					spec.GetValidatorForOpenAPI3Schema(requestBodySchema)
+				if err != nil {
+					return err
+				}
```
brandur-stripe: This is a little subjective, but this code is starting to get indented pretty deeply. Could we maybe introduce a helper function or something?

spec/spec.go
```diff
@@ -2,81 +2,111 @@ package spec
 
 import (
 	"encoding/json"
+	"fmt"
 )
 
+type Components struct {
+	Schemas map[string]*Schema `json:"schemas"`
```
brandur-stripe: Can we give this field a `yaml` tag too?

spec/spec.go
```diff
@@ -2,81 +2,111 @@ package spec
...
+	}
+	for _, supportedField := range supportedFields {
+		delete(rawFields, supportedField)
+	}
+	for unsupportedField := range rawFields {
+		return fmt.Errorf(
+			"unsupported field in JSON schema: '%s'", unsupportedField)
+	}
+
+	type schema2 Schema
```
brandur-stripe: Mind if we find a slightly better name for this? Maybe `schemaAlias`?

Also, I know the pattern is pretty common in stripe-go, but it's still some pretty unintuitive sleight of hand. Mind adding a comment that explains that this is here to avoid a recursive call into this `UnmarshalJSON` function?

spec/spec.go
```diff
@@ -2,81 +2,111 @@ package spec
...
 
-	var rawFields map[string]interface{}
-	err = json.Unmarshal(data, &rawFields)
+	// If you update this, be sure to also update getJSONSchemaForOpenAPI3Schema
+	// to pass the fields through to the JSON validator
+	supportedFields := []string{
+		"anyOf", "enum", "items", "nullable", "properties", "required", "type",
+		"x-expandableFields", "x-expansionResources", "x-resourceId", "$ref",
+		"title", "description",
+	}
```
brandur-stripe: Just OOC, why'd we add this? Just a hedge against future problems?

Also, mind if we move this to a var near the top of the file and put one name per line (makes future diffs cleaner and helps guarantee alphabetization)?

``` go
var supportedFields = []string{
    "anyOf",
    "enum",
    ...
}

File: spec/spec_test.go [hunk 9]
```diff
@@ -1 +1,47 @@
...
+	schema := Schema{
+		Type:     "string",
+		Nullable: true,
+	}
+	v, err := GetValidatorForOpenAPI3Schema(&schema)
+	assert.NoError(t, err)
+	assert.NoError(t, v.Validate("hello"))
+	assert.NoError(t, v.Validate(nil))
+	assert.Error(t, v.Validate(123))
+}
```
brandur-stripe: Thanks for the new tests!

spec/validate.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	return validator, nil
+}
+
+// Given an OpenAPI 3 schema represented as JSON, returns an equivalent JSON
+// Schema represented as JSON. The important difference between OpenAPI 3
+// schemas and JSON schemas is that OpenAPI 3 uses "nullable: true" to mark
+// values that can be null, whereas JSON schemas represent "null" as a type just
+// like "string".
```
brandur-stripe: Thanks for the comment!

spec/validate.go
```diff
@@ -0,0 +1,82 @@
+package spec
+
+import (
+	"github.com/lestrrat/go-jsschema"
+	"github.com/lestrrat/go-jsval"
+	"github.com/lestrrat/go-jsval/builder"
+)
+
+func GetValidatorForOpenAPI3Schema(oaiSchema *Schema) (*jsval.JSVal, error) {
```
brandur-stripe: It would be _super_ great if we could get a basic test for this function (or maybe for `getJSONSchemaForOpenAPI3Schema` where most of the heavy logic lives). Maybe one that manually defines a simplistic spec struct that contains a `Nullable` field and shows how it's converted to an equivalent JSON schema with a `null` type. What do you think?

brandur-stripe:
@tmaxwell-stripe Left a few comments, but amazing work! Let me know how testing it against one of our SDKs goes — I'm super excited to see the results there.

spec/validate.go
tmaxwell-stripe: We have a couple of basic tests in `spec_test.go`. Think we need more? Also, should they be in `schema_test.go` instead?

brandur-stripe: > We have a couple of basic tests in spec_test.go. Think we need more?

Hm, actually would you mind just moving the two `TestValidator_*` tests from `spec_test.go` over to a new `validate_test.go`?

> Also, should they be in schema_test.go instead?

I think it's usually good convention to have `*_test.go` files that are named after a `*.go` file of the same name in the package. IMO it would best for now here to just have two separate test files. You can still easily test the whole package in one command:

``` sh
go test ./spec
```

tmaxwell-stripe: D'oh, I meant to write `validate_test.go` (`validate.go` used to be named `schema.go`). Will do.

brandur:
Thanks for the fixes! LGTM.

brandur-stripe:
@tmaxwell-stripe Should I cut a release for this? Or are we still working out compatibility with stripe-go and stripe-ruby?

tmaxwell-stripe:
`stripe-ruby` should be fully compatible as of the latest version, but I haven't even tried `stripe-go` yet.

brandur-stripe:
> stripe-ruby should be fully compatible as of the latest version, but I haven't even tried stripe-go yet.

Cool, I just tried it and it seems to be doing pretty well. Two failures relating to Bitcoin receivers, but beyond that just one other test to look into.

brandur-stripe:
https://github.com/stripe/stripe-mock/pull/21 was the last fixed we needed to get stripe-go working, so I released this PR and that one as 0.2.0 (now that it's relatively stable, let's try to follow something that looks a little more like semver from now on).

PR #18, opened 2017-09-12
Coerce integer-indexed maps to slices

brandur:
This is unusual for normal "Rack-style" form encoding, but at Stripe on
some endpoints we allow maps that are indexed with integers to be
changed to arrays in the backend. So for example this:

    myarr[0]=foo

Would be interpreted in the API as:

    ["foo"]

The purpose of this is to be able to use arbitrary indexing to update
only certain items, so this also works:

    myarr[3]=foo

Becomes:

    [nil, nil, nil "foo"]

You can find an example of this in the `items` parameter on create
subscription.

Here we augment the coercer to detect if the target type is `array` and
then try to coerce a map to a slice as long as all its indexes were
strings that looked like integers.

I've added a test suite, but have also verified that this fixes a
failing test in my stripe-go PR that's targeting stripe-mock.

r? @tmaxwell-stripe Mind taking a look? Sorry I know this messes with a PR you
have in flight, but hopefully the rebase won't be too bad. We're basically
adding a function and a couple new const/vars.

### DIFF
--- a/param/coercer/coercer_test.go
+++ b/param/coercer/coercer_test.go
@@ -15,7 +15,8 @@ func TestCoerceParams_BooleanCoercion(t *testing.T) {
 		"boolkey": "true",
 	}
 
-	CoerceParams(schema, data)
+	err := CoerceParams(schema, data)
+	assert.NoError(t, err)
 	assert.Equal(t, true, data["boolkey"])
 }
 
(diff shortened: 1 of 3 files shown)


param/coercer/coercer_test.go
```diff
@@ -27,10 +28,75 @@ func TestCoerceParams_IntegerCoercion(t *testing.T) {
...
+		data := map[string]interface{}{
+			"arraykey": map[string]interface{}{
+				"0": "0-index",
+				"2": "2-index",
+			},
+		}
+
+		err := CoerceParams(schema, data)
+		assert.NoError(t, err)
+		assert.Equal(t, "0-index", data["arraykey"].([]interface{})[0])
```
tmaxwell-stripe: Would be nice to also check `2-index`

tmaxwell-stripe:
Other than one minor ask, it looks good to me

param/coercer/coercer_test.go
brandur-stripe: Cool! Added that and also another test that the 1st element is a `nil` as expected.

brandur-stripe:
Pushed change as v0.1.23.

PR #21, opened 2017-09-22
Descend into arrays while doing parameter coercing

brandur-stripe:
This patch adds logic to the coercer so that if it runs into an array,
it can descend into it and coerce its elements according to the JSON
schema defined in `items`.

I found that I needed this to get stripe-go's test suite running with
respect to setting a subscription's items.

r? @tmaxwell-stripe


tmaxwell-stripe:
I think the comment on line 17 is out of date, but otherwise LGTM.

brandur-stripe:
> I think the comment on line 17 is out of date, but otherwise LGTM.

Oops, yep, good call. Fixed and rebased. Thanks for taking a look!

PR #23, opened 2017-10-03
Fix decoded maps when unmarshaling fixtures from a YAML file

### LINKED ISSUE #22: Panic when running with custom fixtures specified
```
$ stripe-mock -version
0.2.0
```

When I run `stripe-mock` with `-fixtures ~/fixtures3.yaml -spec ~/spec3.yaml` the server starts. The fixtures3.yaml and spec3.yaml are from https://raw.githubusercontent.com/stripe/openapi/5f2f0e245c07c918cca185902aceffc64a0b1957/openapi/fixtures3.yaml and https://raw.githubusercontent.com/stripe/openapi/5f2f0e245c07c918cca185902aceffc64a0b1957/openapi/spec3.yaml

But when I do 

```
## Request (3)
curl -X "POST" "http://localhost:12111/v1/sources" \
     -H "Authorization: Bearer sk_test_123" \
     -H "Content-Type: application/x-www-form-urlencoded; charset=utf-8" \
     --data-urlencode "type=bitcoin"
```

the server crashes

```
stripe-mock -fixtures ~/fixtures3.yaml -spec ~/spec3.yaml
Routing to 107 path(s) and 186 endpoint(s) with 104 validator(s)
Listening on port 12111
Request: POST /v1/sources
2017/10/03 10:08:13 http: panic serving 127.0.0.1:51820: Responding to /v1/sources:
Dereferencing '#/components/schemas/source':
Using fixture 'source':
Schema is an object:
{
  "properties": {
    "amount": {
      "nullable": true,
      "type": "integer"
    },
    "client_secret": {
      "type": "string"
    },
    "code_verification": {
      "$ref": "#/components/schemas/source_code_verification_flow"
    },
    "created": {
      "type": "integer"
    },
    "currency": {
      "nullable": true,
      "type": "string"
    },
    "flow": {
      "type": "string"
    },
    "id": {
      "type": "string"
    },
    "livemode": {
      "type": "boolean"
    },
    "metadata": {
      "nullable": true,
      "type": "object"
    },
    "object": {
      "type": "string"
    },
    "owner": {
      "anyOf": [
        {
          "$ref": "#/components/schemas/source_owner"
        }
      ],
      "nullable": true
    },
    "receiver": {
      "$ref": "#/components/schemas/source_receiver_flow"
    },
    "redirect": {
      "$ref": "#/components/schemas/source_redirect_flow"
    },
    "statement_descriptor": {
      "nullable": true,
      "type": "string"
    },
    "status": {
      "type": "string"
    },
    "type": {
      "type": "string"
    },
    "usage": {
      "nullable": true,
      "type": "string"
    }
  },
  "required": [
    "client_secret",
    "created",
    "flow",
    "id",
    "livemode",
    "object",
    "status",
    "type"
  ],
  "type": "object",
  "x-expandableFields": [
    "code_verification",
    "owner",
    "receiver",
    "redirect"
  ],
  "x-resourceId": "source"
}
But example is:
map[owner:map[verified_email:<nil> verified_name:<nil> verified_phone:<nil> address:<nil> email:jenny.rosen@example.com name:<nil> phone:<nil> verified_address:<nil>] redirect:map[url: return_url: status:] statement_descriptor: usage:single_use code_verification:map[attempts_remaining:%!s(int=0) status:] id:src_1B3xSU2eZvKYlo2CYdL110Ll object:source type:bitcoin client_secret:src_client_secret_BQp65YD6eXk9NLfGKOwigCoK created:%!s(int=1234567890) flow:receiver amount:%!s(int=1000) receiver:map[address:test_1MBhWS3uv4ynCfQXF3xQjJkzFPukr4K56N amount_charged:%!s(int=0) amount_received:%!s(int=0) amount_returned:%!s(int=0) refund_attributes_method:email refund_attributes_status:missing] status:pending currency:usd livemode:%!s(bool=false) metadata:map[]]
goroutine 20 [running]:
net/http.(*conn).serve.func1(0xc420dce8c0)
	/usr/local/Cellar/go/1.9/libexec/src/net/http/server.go:1697 +0xd0
panic(0x134df00, 0xc420900960)
	/usr/local/Cellar/go/1.9/libexec/src/runtime/panic.go:491 +0x283
main.(*DataGenerator).generateInternal(0xc42087fb00, 0xc420ae4780, 0xc42027c365, 0xb, 0x0, 0xc420900160, 0xc420076660, 0x60, 0xc420a69860, 0xc420a69930, ...)
	/Users/brandur/Documents/go/src/github.com/stripe/stripe-mock/generator.go:151 +0x16af
main.(*DataGenerator).Generate(0xc42087fb00, 0xc420ae4780, 0xc42027c365, 0xb, 0x0, 0x1, 0x0, 0x0, 0x0)
	/Users/brandur/Documents/go/src/github.com/stripe/stripe-mock/generator.go:22 +0x120
main.(*StubServer).HandleRequest(0xc420b5d9c0, 0x159b240, 0xc4208fc540, 0xc4203e2000)
	/Users/brandur/Documents/go/src/github.com/stripe/stripe-mock/server.go:175 +0x55b
main.(*StubServer).HandleRequest-fm(0x159b240, 0xc4208fc540, 0xc4203e2000)
	/Users/brandur/Documents/go/src/github.com/stripe/stripe-mock/main.go:75 +0x48
net/http.HandlerFunc.ServeHTTP(0xc4208f9f50, 0x159b240, 0xc4208fc540, 0xc4203e2000)
	/usr/local/Cellar/go/1.9/libexec/src/net/http/server.go:1918 +0x44
net/http.(*ServeMux).ServeHTTP(0x15e7c60, 0x159b240, 0xc4208fc540, 0xc4203e2000)
	/usr/local/Cellar/go/1.9/libexec/src/net/http/server.go:2254 +0x130
net/http.serverHandler.ServeHTTP(0xc42046fd40, 0x159b240, 0xc4208fc540, 0xc4203e2000)
	/usr/local/Cellar/go/1.9/libexec/src/net/http/server.go:2619 +0xb4
net/http.(*conn).serve(0xc420dce8c0, 0x159b8c0, 0xc42084a980)
	/usr/local/Cellar/go/1.9/libexec/src/net/http/server.go:1801 +0x71d
created by net/http.(*Server).Serve
	/usr/local/Cellar/go/1.9/libexec/src/net/http/server.go:2720 +0x288
```

I didn't make any changes to the files, so I'm not sure what is going on.

When I remove the `-fixture path.yaml` and only run it with `-spec path.yaml` it works fine.

Any idea?

brandur-stripe:
As shown in #22, when we try to unmarshal fixtures from a YAML file, we
end up with a crash in the generator due to a type mismatch.

This turns out to be because of [1]. Because YAML technically supports
boolean keys, the YAML library unmarshals to maps in a way to handle
them (`map[interface{}]interface{}`), even if it's not suitable for most
users (the issue strongly suggests that a lot of people would like
`map[string]interface{}`).

Here we work around the problem by post-processing fixtures that we've
loaded via YAML by recursing through and changing all instances of
`map[interface{}]interface{}` to `map[string]interface{}`. I've added a
few basic tests, but also tested that the call reported in #22 now works
as expected.

Fixes #22.

[1] https://github.com/go-yaml/yaml/issues/139

r? @tmaxwell-stripe

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"reflect"
 	"sort"
 	"strings"
 
@@ -148,8 +149,9 @@ func (g *DataGenerator) generateInternal(schema *spec.Schema, requestPath string
 	if schema.Type == "object" && schema.Properties != nil {
 		exampleMap, ok := example.value.(map[string]interface{})
 		if !ok {
-			panic(fmt.Sprintf("%sSchema is an object:\n%s\nBut example is:\n%s",
-				context, schema, example.value))
+			panic(fmt.Sprintf(
+				"%sSchema is an object:\n%s\n\nBut example is (type: %v):\n%s",
+				context, schema, reflect.TypeOf(example.value), example.value))
 		}
 
 		resultMap := make(map[string]interface{})
--- a/mapstr.go
+++ b/mapstr.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"fmt"
+)
+
+// stringifyKeysMapValue recurses into v and changes all instances of
+// map[interface{}]interface{} to map[string]interface{}. This is useful to
+// work around the impedence mismatch between JSON and YAML unmarshaling that's
+// described here:
+//
+// https://github.com/go-yaml/yaml/issues/139
+func stringifyKeysMapValue(v interface{}) interface{} {
+	switch v := v.(type) {
+	case []interface{}:
+		return stringifyKeysInterfaceArray(v)
+	case map[interface{}]interface{}:
+		return stringifyKeysInterfaceMap(v)
+	default:
+		return v
+	}
+}
+
+//
+// helpers
+//
+
+func stringifyKeysInterfaceArray(in []interface{}) []interface{} {
+	res := make([]interface{}, len(in))
+	for i, v := range in {
+		res[i] = stringifyKeysMapValue(v)
+	}
+	return res
+}
+
+func stringifyKeysInterfaceMap(in map[interface{}]interface{}) map[string]interface{} {
+	res := make(map[string]interface{})
+	for k, v := range in {
+		res[fmt.Sprintf("%v", k)] = stringifyKeysMapValue(v)
+	}
+	return res
+}
--- a/mapstr.go
+++ b/mapstr.go
@@ -36,7 +36,11 @@ func stringifyKeysInterfaceArray(in []interface{}) []interface{} {
 func stringifyKeysInterfaceMap(in map[interface{}]interface{}) map[string]interface{} {
 	res := make(map[string]interface{})
 	for k, v := range in {
-		res[fmt.Sprintf("%v", k)] = stringifyKeysMapValue(v)
+		kStr, ok := k.(string)
+		if !ok {
+			panic(fmt.Sprintf("Expected all keys in maps to be strings. Got: %v", k))
+		}
+		res[kStr] = stringifyKeysMapValue(v)
 	}
 	return res
 }
--- a/main.go
+++ b/main.go
@@ -114,6 +114,15 @@ func getFixtures(fixturesPath string) (*spec.Fixtures, error) {
 
 	if isYAML {
 		err = yaml.Unmarshal(data, &fixtures)
+		if err == nil {
+			// To support boolean keys, the `yaml` package unmarshals maps to
+			// map[interface{}]interface{}. Here we recurse through the result
+			// and change all maps to map[string]interface{} like we would've
+			// gotten from `json`.
+			for k, v := range fixtures.Resources {
+				fixtures.Resources[k] = stringifyKeysMapValue(v)
+			}
+		}
 	} else {
 		err = json.Unmarshal(data, &fixtures)
 	}
--- a/mapstr_test.go
+++ b/mapstr_test.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+)
+
+func TestStringifyKeysMapValue(t *testing.T) {
+	assert.Equal(t,
+		map[string]interface{}{
+			"arrkey": []interface{}{
+				123,
+				map[string]interface{}{
+					"intkey": 123,
+				},
+			},
+			"boolkey": true,
+			"intkey":  123,
+			"mapkey": map[string]interface{}{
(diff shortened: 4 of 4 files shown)


generator.go
```diff
@@ -148,8 +149,9 @@ func (g *DataGenerator) generateInternal(schema *spec.Schema, requestPath string
 	if schema.Type == "object" && schema.Properties != nil {
 		exampleMap, ok := example.value.(map[string]interface{})
 		if !ok {
-			panic(fmt.Sprintf("%sSchema is an object:\n%s\nBut example is:\n%s",
-				context, schema, example.value))
+			panic(fmt.Sprintf(
+				"%sSchema is an object:\n%s\n\nBut example is (type: %v):\n%s",
+				context, schema, reflect.TypeOf(example.value), example.value))
```
brandur-stripe: Also added a type hint here to help with debugging in the future (it's how I found this problem).

mapstr.go
```diff
@@ -0,0 +1,42 @@
...
+	for i, v := range in {
+		res[i] = stringifyKeysMapValue(v)
+	}
+	return res
+}
+
+func stringifyKeysInterfaceMap(in map[interface{}]interface{}) map[string]interface{} {
+	res := make(map[string]interface{})
+	for k, v := range in {
+		res[fmt.Sprintf("%v", k)] = stringifyKeysMapValue(v)
```
tmaxwell-stripe: I think we should check that the key is already a string and raise an error if it isn't. Otherwise, this can lead to some really strange behavior. Suppose the original YAML file looks like:
```
in: a
on: b
under: c
```
The YAML parser will interpret `in` and `under` as strings but it will interpret `on` as the boolean `true`. So the output of `stringifyKeysMapValue` will be:
```
map[string]interface{}{
  "in": "a",
  "true": "b",
  "under": "c",
}
```
which is super confusing and almost certainly not what was intended.

tmaxwell-stripe:
Thanks for tracking down that bug! I left a comment suggesting a slightly different behavior, but otherwise it looks good.

mapstr.go
brandur-stripe: Great point. I adapted this code from elsewhere, but yep, in our case I think it makes perfect sense to be as loud as possible about this problem because we expect it to _never_ happen.

Patched in 24a2bb4. Thanks for taking a look.

tmaxwell-stripe:
LGTM, thanks!

PR #28, opened 2017-10-31
Update OpenAPI specification

brandur-stripe:
This should include a few updates including the exchange rates APIs.


brandur-stripe:
FYI @tmaxwell-stripe.

Pulling this in.

PR #32, opened 2017-11-12
Define a StripeError struct and use this on all API errors

remi-stripe:
This lets stripe-mock return API errors as a JSON payload instead of just the error message.

This fixes https://github.com/stripe/stripe-mock/issues/31 and lets us use stripe-mock better in stripe-python tests for https://github.com/stripe/stripe-python/pull/367

It also ensures that `Request-Id` is properly returned as a header on all requests except the authentication error ones.

### DIFF
--- a/server.go
+++ b/server.go
@@ -114,6 +114,9 @@ func (s *StubServer) HandleRequest(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Every response needs a Request-Id header except the invalid authorization
+	w.Header().Set("Request-Id", "req_123")
+
 	route := s.routeRequest(r)
 	if route == nil {
 		stripeError.ErrorInfo.Message = fmt.Sprintf(invalidRoute,
--- a/server.go
+++ b/server.go
@@ -122,6 +122,7 @@ func (s *StubServer) HandleRequest(w http.ResponseWriter, r *http.Request) {
 
 	response, ok := route.operation.Responses["200"]
 	if !ok {
+		fmt.Printf("Couldn't find 200 response in spec\n")
 		writeResponse(w, r, start, http.StatusInternalServerError,
 			createInternalServerError())
 		return
--- a/server.go
+++ b/server.go
@@ -26,6 +26,8 @@ const (
 	invalidRoute = "Unrecognized request URL (%s: %s)."
 
 	internalServerError = "An internal error occurred."
+
+	typeInvalidRequestError = "invalid_request_error"
 )
 
 type ErrorInfo struct {
@@ -99,17 +101,10 @@ func (s *StubServer) HandleRequest(w http.ResponseWriter, r *http.Request) {
 	start := time.Now()
 	fmt.Printf("Request: %v %v\n", r.Method, r.URL.Path)
 
-	// Prepare internal error as it's the most commonly used here
-	stripeError := &ResponseError{
-		ErrorInfo: ErrorInfo{
-			Message: internalServerError,
-			Type:    "invalid_request_error",
-		},
-	}
-
 	auth := r.Header.Get("Authorization")
 	if !validateAuth(auth) {
-		stripeError.ErrorInfo.Message = fmt.Sprintf(invalidAuthorization, auth)
+		message := fmt.Sprintf(invalidAuthorization, auth)
+		stripeError := createStripeError(typeInvalidRequestError, message)
 		writeResponse(w, r, start, http.StatusUnauthorized, stripeError)
 		return
 	}
@@ -119,22 +114,23 @@ func (s *StubServer) HandleRequest(w http.ResponseWriter, r *http.Request) {
 
 	route := s.routeRequest(r)
 	if route == nil {
-		stripeError.ErrorInfo.Message = fmt.Sprintf(invalidRoute,
-			r.Method, r.URL.Path)
+		message := fmt.Sprintf(invalidRoute, r.Method, r.URL.Path)
+		stripeError := createStripeError(typeInvalidRequestError, message)
 		writeResponse(w, r, start, http.StatusNotFound, stripeError)
 		return
 	}
 
 	response, ok := route.operation.Responses["200"]
 	if !ok {
-		fmt.Printf("Couldn't find 200 response in spec\n")
-		writeResponse(w, r, start, http.StatusInternalServerError, stripeError)
+		writeResponse(w, r, start, http.StatusInternalServerError,
+			createInternalServerError())
 		return
 	}
 	responseContent, ok := response.Content["application/json"]
 	if !ok || responseContent.Schema == nil {
 		fmt.Printf("Couldn't find application/json in response\n")
-		writeResponse(w, r, start, http.StatusInternalServerError, stripeError)
+		writeResponse(w, r, start, http.StatusInternalServerError,
+			createInternalServerError())
 		return
 	}
 
@@ -149,7 +145,8 @@ func (s *StubServer) HandleRequest(w http.ResponseWriter, r *http.Request) {
 		formBytes, err := ioutil.ReadAll(r.Body)
 		if err != nil {
 			fmt.Printf("Couldn't read request body: %v\n", err)
-			writeResponse(w, r, start, http.StatusInternalServerError, stripeError)
+			writeResponse(w, r, start, http.StatusInternalServerError,
+				createInternalServerError())
 			return
 		}
 		r.Body.Close()
@@ -158,7 +155,8 @@ func (s *StubServer) HandleRequest(w http.ResponseWriter, r *http.Request) {
 	requestData, err := parser.ParseFormString(formString)
 	if err != nil {
 		fmt.Printf("Couldn't parse query/body: %v\n", err)
-		writeResponse(w, r, start, http.StatusInternalServerError, stripeError)
+		writeResponse(w, r, start, http.StatusInternalServerError,
+			createInternalServerError())
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -86,29 +99,42 @@ func (s *StubServer) HandleRequest(w http.ResponseWriter, r *http.Request) {
 	start := time.Now()
 	fmt.Printf("Request: %v %v\n", r.Method, r.URL.Path)
 
+	// Prepare internal error as it's the most commonly used here
+	stripeError := &ResponseError{
```
tmaxwell-stripe: Instead of preparing this as a variable, can we instead have a function that prepares an error response and returns it? I think that would be slightly clearer.

server.go
```diff
@@ -86,29 +99,42 @@ func (s *StubServer) HandleRequest(w http.ResponseWriter, r *http.Request) {
...
-		writeResponse(w, r, start, http.StatusNotFound, nil)
+		stripeError.ErrorInfo.Message = fmt.Sprintf(invalidRoute,
+			r.Method, r.URL.Path)
+		writeResponse(w, r, start, http.StatusNotFound, stripeError)
 		return
 	}
 
 	response, ok := route.operation.Responses["200"]
 	if !ok {
 		fmt.Printf("Couldn't find 200 response in spec\n")
```
tmaxwell-stripe: Would it make sense to have one function that both prints the error message and generates the error response?

server_test.go
```diff
@@ -16,69 +16,58 @@ import (
...
-	err = json.Unmarshal(body, &data)
+	err := json.Unmarshal(body, &data)
 	assert.NoError(t, err)
 	_, ok := data["id"]
 	assert.True(t, ok)
-}
 
-func TestStubServer_SetsSpecialHeaders(t *testing.T) {
-	server := getStubServer(t)
+	// Ensure error response has the correct shape
```
tmaxwell-stripe: Maybe this should be in a separate test?

server_test.go
```diff
@@ -275,3 +264,27 @@ func getStubServer(t *testing.T) *StubServer {
...
 	return server
 }
+
+func getDefaultHeaders() map[string]string {
+	headers := make(map[string]string)
+	headers["Authorization"] = "Bearer sk_test_123"
+	return headers
+}
+
+func sendRequest(t *testing.T, method string, url string, params string,
```
tmaxwell-stripe: Defining this helper function was a really good idea. Thank you!

server.go
```diff
@@ -132,7 +158,7 @@ func (s *StubServer) HandleRequest(w http.ResponseWriter, r *http.Request) {
 	requestData, err := parser.ParseFormString(formString)
 	if err != nil {
 		fmt.Printf("Couldn't parse query/body: %v\n", err)
-		writeResponse(w, r, start, http.StatusInternalServerError, nil)
+		writeResponse(w, r, start, http.StatusInternalServerError, stripeError)
```
tmaxwell-stripe: If the form string is invalid, shouldn't that be a 400 instead of a 500? (But it was like that before, so don't feel obligated to fix in this PR.)

tmaxwell-stripe:
Looks good overall. I suggested a couple of minor changes.

server.go
remi-stripe: @tmaxwell-stripe fixed in https://github.com/stripe/stripe-mock/pull/32/commits/f9ad1e76482ce44b1eb9367cc8485dac93dd39a1 is this what you had in mind?

server.go
remi-stripe: Made it better in https://github.com/stripe/stripe-mock/pull/32/commits/f9ad1e76482ce44b1eb9367cc8485dac93dd39a1 though the message is still generate first and then passed as a parameter. LMK if that fits the ask

server_test.go
remi-stripe: https://github.com/stripe/stripe-mock/pull/32/commits/e5dfd59b237c2cbdf520d0e3c3b784739cad5969

remi-stripe:
@tmaxwell-stripe made some fixes PTAL

Also Go is not a language I know beyond making changes work. Please let me know if some things could be done in a better way. I'm happy to spend more time reworking this.

server.go
tmaxwell-stripe: Yep, that's it. Thanks!

server.go
remi-stripe: The issue is that the strings can have 0/1/N dynamic parameters (a bit like printf) so we'd have to rebuild a printf-like version right?

Something like `createStripeError(errorType string, errorMessagePattern string, params ...string)`.

Do you think it's worth it? Happy to implement if so!

tmaxwell-stripe: Another option would be something like `createInternalServerError(fmt.Sprintf("missing widget: %s", widget))`.

I don't have a strong opinion either way; I just wanted to raise the possibility. I'll leave it up to your judgement.

remi-stripe: Ah yes I did that at first but with the rule of rows no more than 80 characters then it ended up spanning 2 rows anyway.

For example:
```
	if route == nil {
		message := fmt.Sprintf(invalidRoute, r.Method, r.URL.Path)
		stripeError := createStripeError(typeInvalidRequestError, message)
		writeResponse(w, r, start, http.StatusNotFound, stripeError)
		return
	}
```
would be
```
	if route == nil {
		message := 
		stripeError := createStripeError(typeInvalidRequestError,
			fmt.Sprintf(invalidRoute, r.Method, r.URL.Path))
		writeResponse(w, r, start, http.StatusNotFound, stripeError)
		return
	}
```

Do you prefer the latter?

tmaxwell-stripe: Either is fine; I'll leave it to your judgement.

server.go
```diff
@@ -88,27 +103,34 @@ func (s *StubServer) HandleRequest(w http.ResponseWriter, r *http.Request) {
...
-		writeResponse(w, r, start, http.StatusNotFound, nil)
+		message := fmt.Sprintf(invalidRoute, r.Method, r.URL.Path)
+		stripeError := createStripeError(typeInvalidRequestError, message)
+		writeResponse(w, r, start, http.StatusNotFound, stripeError)
 		return
 	}
 
 	response, ok := route.operation.Responses["200"]
 	if !ok {
-		fmt.Printf("Couldn't find 200 response in spec\n")
```
tmaxwell-stripe: Shouldn't we keep this `fmt.Printf()` call?

server.go
remi-stripe: Cool, I'll leave as is then because that's my personal preference but we can always improve this in a future PR if needed.

server.go
remi-stripe: we should sorry I missed this one when reviewing my own diff, fixed.

tmaxwell-stripe:
LGTM

remi-stripe:
@tmaxwell-stripe Can I leave to you to merge and deploy the new version? I don't think I have the right permissions for this. Once it's live I can use this in stripe-python!

server_test.go
```diff
@@ -16,69 +16,57 @@ import (
...
+	assert.Equal(t, http.StatusUnauthorized, resp.StatusCode)
 
-	// Does this regardless of endpoint
-	req := httptest.NewRequest("GET", "https://stripe.com/", nil)
-	w := httptest.NewRecorder()
-	server.HandleRequest(w, req)
+	var data map[string]interface{}
+	err := json.Unmarshal(body, &data)
+	assert.NoError(t, err)
+	errorInfo, present := data["error"].(map[string]interface{})
```
brandur-stripe: Mind renaming `present` to `ok` here? It's a pretty common convention in Go.

server.go
```diff
@@ -405,3 +433,20 @@ func validateAuth(auth string) bool {
...
+// This creates a Stripe error to return in case of API errors.
+func createStripeError(errorType string, errorMessage string) *ResponseError {
+	stripeError := &ResponseError{
+		ErrorInfo: ErrorInfo{
+			Message: errorMessage,
+			Type:    errorType,
+		},
+	}
+
+	return stripeError
```
brandur-stripe: Can we do away with assigning `stripeError` here and just return `&ResponseError` directly?

server.go
```diff
@@ -22,8 +22,23 @@ const (
...
+	typeInvalidRequestError = "invalid_request_error"
 )
 
+type ErrorInfo struct {
+	Type    string `json:"type"`
+	Message string `json:"message"`
+}
+
+type ResponseError struct {
+	ErrorInfo ErrorInfo `json:"error"`
```
brandur-stripe: Given that `ErrorInfo` never needs to be addressed directly, this is probably a good place to use an anonymous struct like:

``` go
type ResponseError struct {
    ErrorInfo struct {
        ...
    } `json:"error"`
}
```

Not too big of a deal though.

remi-stripe: @brandur-stripe okay so I fixed this one but I had to repeat the struct definition in [createStripeError()](https://github.com/stripe/stripe-mock/pull/32/commits/23e9582c430ebe0bac4248252043a1c2a5136a38#diff-34c6b408d72845d076d47126c29948d1R441) which feels potentially less clean (though maybe more idiomatic)

server.go
remi-stripe: fixed

server_test.go
remi-stripe: fixed

remi-stripe:
@brandur-stripe  PTAL

server.go
brandur-stripe: Yeah, that's an unfortunate side effect. That said, I think it works well as you have it now. Thanks!

brandur-stripe:
Thanks for taking a look at those comments Remi! Shipping this out.

brandur-stripe:
Released as v0.5.0.

PR #38, opened 2017-12-14
build and run with docker

MattMasters:


brandur-stripe:
Cool. Thanks!

Is it common to have a CI-based process just to make sure that the `Dockerfile` builds as expected?

MattMasters:
No problem! 

Yeah if you can plug that into a CI build you could bring the container image down significantly, by building from the ‘scratch’ docker image and cross compiling the go binary for that target.

You could also push that to docked hub and users wouldn’t have to checkout the project and build the docket image locally. 

Hope that makes sense?

All the best,
Matt

brandur-stripe:
Nice — actually I was more wondering this: is there anything that we add to our existing CI build to help ensure that this `Dockerfile` works? Even if it does today, it feels like when future changes are made we wouldn't be doing a good job of testing for regressions and someone could easily break it.

MattMasters:
Hey, sorry for the delay.

I've updated this PR to build the Docker container in your current Travis run, and then exec onto the running container and run the go tests. Following this, I would suggest that you push the tested image to docker hub [example here](https://sebest.github.io/post/using-travis-ci-to-build-docker-images/).

Cheers,
Matt

brandur-stripe:
@MattMasters Nice work!

Two more minor things:

1. Can you move the Docker CI commands into the main `script` section after the existing commands in there? If we have a test failure, I'd rather see it directly rather than through the Docker build.
2. Any particular reason to use the Alpine Docker image? Would it be better to lock to one of the existing version-based ones like 1.9?

> Following this, I would suggest that you push the tested image to docker hub example here.

Thanks for the advice! Very helpful. I might wait until someone asks for it, but this is great to know.

MattMasters:
I'm using alpine instead of the 1.9 base image because of the size, the apline images are 83MB whereas the 1.9 one is 272MB. Trying to limit the time the build is sat downloading docker images :)

brandur-stripe:
> I'm using alpine instead of the 1.9 base image because of the size, the apline images are 83MB whereas the 1.9 one is 272MB. Trying to limit the time the build is sat downloading docker images :)

Hah, that is a huge difference. Alright, thanks a lot for explaining and for the extra work here!

PR #39, opened 2018-01-16
Use fixtures and spec from jhoon-stripe/openapi

jhoon-stripe:
This PR provides a version of `stripe-mock` against which we can run Stripe language binding tests that include test cases against an upcoming Stripe API version with separate products and plans.

cc @brandur


jhoon-stripe:
We'd actually need to cut a release of stripe-mock in order to use this is in stripe-go CI, so I'm going  close out this PR, at least for now.

PR #40, opened 2018-01-25
Update OpenAPI and fixtures

brandur-stripe:
Periodic update for the bundled OpenAPI specification and fixtures in
use by `stripe-mock`.

cc @stripe/api-libraries


brandur-stripe:
Released as 0.6.0.

PR #45, opened 2018-02-16
Always print the stripe-mock version

jkakar-stripe:
I've found it a bit confusing, during stripe-mock version changes, to know what version I'm running. I discovered that I can run `stripe-mock --version` to see what version I have, but it'd be nice if I never had to do that. This (maybe contentious) change will make stripe-mock emit its version each time it runs.

r? @stripe/api-libraries

### DIFF
--- a/main.go
+++ b/main.go
@@ -43,8 +43,8 @@ func main() {
 	flag.BoolVar(&showVersion, "version", false, "Show version")
 	flag.Parse()
 
+	fmt.Printf("stripe-mock %s\n", version)
 	if showVersion || len(flag.Args()) == 1 && flag.Arg(0) == "version" {
-		fmt.Printf("%s\n", version)
 		return
 	}
 


brandur-stripe:
Given that the program is relatively verbose already (even without `--verbose`), I think this would be okay.

It's weird having that empty conditional that no longer does anything though (the check on `--version`). What do you think we should do about that?

jkakar-stripe:
@brandur Yeah, I agree that it looks a bit odd, but my thought was that if you pass `--version` you want the program to quit right after it emits the version info, so that's why I left that there.

brandur-stripe:
> @brandur Yeah, I agree that it looks a bit odd, but my thought was that if you pass --version you want the program to quit right after it emits the version info, so that's why I left that there.

Yeah, that makes sense. Cool, bringing this in!

PR #47, opened 2018-03-07
New format for list resource URLs in spec

tmaxwell-stripe:
Previously, list resource URLs were described in the spec as `"enum": ["/v1/charges"]`. This wasn't correct because if a filter was applied to the list, then the URL would actually be something of the form `/v1/charges?...`. The spec was recently updated to use `"pattern": "^/v1/charges"`. This PR updates `stripe-mock` to correctly interpret the new format.

I also added support for non-string values in the `enum` field, because we now use that in a few places.


brandur-stripe:
LGTM.

Thanks for taking a crack at this Tim.

Would you mind squashing your commits while bringing this in?

PR #50, opened 2018-03-21
Push Docker image to Docker Hub

### LINKED ISSUE #49: Publish docker image to docker hub?
Is there an official build of the docker image that can be used without building? I couldn't find one on docker hub.

brandur-stripe:
Pushes the built Docker image to Docker Hub as part of a Travis deploy
step. This allows Docker users to access the image more easily.

There are a few caveats with this implementation:

1. We're only building `master`. If this ends up being something we use,
   we should build tagged version images as well.

2. It pushes to `stripemock/stripe-mock` instead of `stripe/stripe-mock`
   because we're having a little trouble getting the collaboration
   privileges worked out on Docker Hub's Stripe organization right now.
   We should change it to `stripe/stripe-mock` when possible.

Fixes #49.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,12 +19,45 @@ script:
   - make vet
   - scripts/check_gofmt.sh
 
-  # Build and test the Docker image
-  - docker build -t stripe-mock .
-  - docker run -d --name stripe-mock -p 12111:12111 stripe-mock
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Alright, let's see if this works ...

ob-stripe:
Nice!

PR #53, opened 2018-03-30
Fall back to the `PORT` environmental variable

### LINKED ISSUE #42: Deployment (Heroku button)
Hi,
this is a suggestion but it would be great to be able to deploy an instance through the click of a button 

https://devcenter.heroku.com/articles/heroku-button

Thanks.

brandur-stripe:
Adds logic to fall back to the `PORT` environmental variable if one was
not specified with `--port`. This makes deployment more flexible on some
platforms like Heroku.

The precedence is now (in order of most preferred to least):

1. `--port` via command line arguments.
2. `PORT` via environment.
3. Default port (`12111`).

Here it is in action:

    $ PORT=12345 $GOPATH/bin/stripe-mock
    stripe-mock master
    Routing to 115 path(s) and 196 endpoint(s) with 108 validator(s)
    Listening on port 12345

Fixes #42.

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,7 +35,7 @@ func main() {
 	var specPath string
 	var unix string
 
-	flag.IntVar(&port, "port", 0, "Port to listen on")
+	flag.IntVar(&port, "port", 0, "Port to listen on (also respects PORT from environment)")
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Thanks OB.

PR #54, opened 2018-03-30
Send valid path to generator while running test suite

brandur-stripe:
So I couldn't figure out what exactly changed here, but as reported in
our channel yesterday, `stripe-mock`'s test suite fails on the latest
version of the OpenAPI spec.

The reason is that when generating list resources, we're currently
filling in `url` with the dud string `<test request path>`, and doing so
causes the generated resource to fail to validate because that string is
not in the OpenAPI spec's `enum` set for `url`.

Here we resolve the problem by plumbing through each URL's path from the
ingested OpenAPI spec and sending it onto the generator.

This doesn't quite fix the problems with the latest OpenAPI though —
there's at least one more case of an invalid fixture which I'm trying to
resolve separately.

r? @tmaxwell-stripe
cc @alexander-stripe
cc @stripe/api-libraries

### DIFF
--- a/generator_test.go
+++ b/generator_test.go
@@ -177,7 +177,7 @@ func TestValidFixtures(t *testing.T) {
 
 // Tests that DataGenerator can generate an example of the given schema, and
 // that the example validates against the schema correctly
-func testCanGenerate(t *testing.T, schema *spec.Schema, expand bool) {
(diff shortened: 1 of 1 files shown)


tmaxwell-stripe:
LGTM

brandur-stripe:
Thanks Tim.

PR #55, opened 2018-03-30
Update bundled OpenAPI spec

brandur-stripe:
Updates the bundled OpenAPI specification. Notably, this includes a fix
to the token fixture which was previously not validating against its
ostensible schema.


brandur-stripe:
Had to do one more round of fixes server-side, but that seems to have done it! Merging.

brandur-stripe:
Released as 0.10.0.

PR #59, opened 2018-04-05
Reflect an object's primary ID from a path back in the response

brandur-stripe:
When requesting a path from `stripe-mock` with an object's primary ID in
the URL:

```
GET /v1/charges/ch_idFromURL
```

Previously, we'd return the ID from the bundled fixtures. This patch
changes the behavior to try and detect when the path contains an
object's primary ID and reflect that back into the respond instead:

```
{
    "id": "ch_idFromURL",
}
```

(Previously, this would have been `ch_123` or whatever the fixtures
contained.)

We also tweak things in a few other places for consistency. If we find a
string value in our generated response with the same value as the ID we
replaced, we replace it as well. This allows an embedded resource with
an ID field that references its parent to reference it with a sane ID
(e.g., the `charge` field on a refund that's embedded in a charge's
`refunds` sublist).

We also modify URLs in embedded sublists to contain the right ID. e.g.,

```
{
    "url": "/v1/charges/ch_idFromURL"
}
```

This is necessary because those URLs are used by our client libraries to
determine where they should make calls to (e.g., in `stripe-ruby`
calling `charge.refunds.create(...)` uses that `url` values to know
where to post).

All of this is designed to get our test suite working again with the
latest OpenAPI fixtures. In particular, this will allow us to return the
right `id` for any deleted object, which will allow `stripe-ruby`'s web
mock checks to pass correctly.

r? @tmaxwell
cc @stripe/api-libraries

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -19,8 +19,8 @@ type DataGenerator struct {
 }
 
 // Generate generates a fixture response.
-func (g *DataGenerator) Generate(schema *spec.Schema, requestPath string, expansions *ExpansionLevel) (interface{}, error) {
-	return g.generateInternal(schema, requestPath, expansions, nil, fmt.Sprintf("Responding to %s:\n", requestPath))
+func (g *DataGenerator) Generate(schema *spec.Schema, requestPath string, id *string, expansions *ExpansionLevel) (interface{}, error) {
+	return g.generateInternal(schema, requestPath, id, nil, true, expansions, nil, fmt.Sprintf("Responding to %s:\n", requestPath))
 }
 
 // The purpose of this simple wrapper is so that we can write "example = nil"
@@ -30,7 +30,7 @@ type Example struct {
 	value interface{}
 }
 
-func (g *DataGenerator) generateInternal(schema *spec.Schema, requestPath string, expansions *ExpansionLevel, example *Example, context string) (interface{}, error) {
+func (g *DataGenerator) generateInternal(schema *spec.Schema, requestPath string, id *string, replacedID *string, doReplaceID bool, expansions *ExpansionLevel, example *Example, context string) (interface{}, error) {
 	// This is a bit of a mess. We don't have an elegant fully-general approach to
 	// generating examples, just a bunch of specific cases that we know how to
 	// handle. If we find ourselves in a situation that doesn't match any of the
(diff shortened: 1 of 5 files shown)


generator.go
```diff
@@ -30,7 +30,7 @@ type Example struct {
 	value interface{}
 }
 
-func (g *DataGenerator) generateInternal(schema *spec.Schema, requestPath string, expansions *ExpansionLevel, example *Example, context string) (interface{}, error) {
+func (g *DataGenerator) generateInternal(schema *spec.Schema, requestPath string, id *string, replacedID *string, doReplaceID bool, expansions *ExpansionLevel, example *Example, context string) (interface{}, error) {
```
brandur-stripe: Not super happy about continued argument proliferation here. I think we need to change this to an args struct, but I want to do that on a different PR.

brandur-stripe:
@tmaxwell-stripe Bump on this one — I suspect the notification might've been lost somewhere. Thanks!

generator.go
```diff
@@ -129,6 +133,20 @@ func (g *DataGenerator) generateInternal(schema *spec.Schema, requestPath string
...
+	//
+	// For example, a charge may contain a sublist of refunds. If we replaced
+	// the charge's ID, we also want to replace that charge ID in every one of
+	// the child refunds.
+	if replacedID != nil && schema.Type == "string" {
+		valStr, ok := example.value.(string)
+		if ok && valStr == *replacedID {
+			example = &Example{value: *id}
+		}
+	}
```
tmaxwell-stripe: What if we did this calculation in the fixture generation process instead? For example, we could set the `id` field's value in the fixture to something like `$id`, and then `stripe-mock` could just replace the string `$id` with the actual ID wherever it appears. I think this would be more robust and flexible.

brandur-stripe: Hm, definitely an interesting idea, but wouldn't we just be moving this logic from one place to another? I'm not sure that doing it from within the API's generator will make the process all that much easier.

I'm also kind of a fan that today the fixtures really are just fixtures — i.e., we haven't developed any specialized pseudo-format that a consumer needs to worry about pre-processing before they're usable. I don't feel quite strongly enough about that to say that we should never do it, but I'd generally prefer to delay until we're really sure that it's something we want.

generator.go
```diff
@@ -19,8 +19,8 @@ type DataGenerator struct {
 }
 
 // Generate generates a fixture response.
-func (g *DataGenerator) Generate(schema *spec.Schema, requestPath string, expansions *ExpansionLevel) (interface{}, error) {
-	return g.generateInternal(schema, requestPath, expansions, nil, fmt.Sprintf("Responding to %s:\n", requestPath))
+func (g *DataGenerator) Generate(schema *spec.Schema, requestPath string, id *string, expansions *ExpansionLevel) (interface{}, error) {
```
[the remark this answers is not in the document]
brandur-stripe: Hm, good point. Do you mind if I punt that feature for now? I plan on doing a refactoring pass after this is in (sooner is better than later just so we can get that stripe-ruby PR merged), and could work on this improvement after that's in.

brandur-stripe:
Thanks for the feedback Tim!

ptal @tmaxwell-stripe

brandur-stripe:
Hey @tmaxwell-stripe, one more bump on this one — I wouldn't normally be insistent, but this is blocking support for the flexible billing production in `stripe-ruby`. Are you okay with me putting this in mostly as-is? I think we can proceed with some other improvements afterwards.

tmaxwell-stripe:
Yeah, I think this is fine for now.

LGTM

brandur-stripe:
TY Tim. I'm going to start on some more fixes/improvements immediately.

PR #66, opened 2018-04-26
Bring in OpenAPI including "create file" endpoint

brandur-stripe:
Here we bring in a fresh version of OpenAPI that includes the "create
file" endpoint. This should bring us to a place where it's possible to
test a complete file upload against `stripe-mock`.


brandur-stripe:
Oops, this ended up coming in elsewhere.

PR #67, opened 2018-04-26
Support primitives in an `anyOf` from the parameters coercer

brandur-stripe:
I was running `stripe-ruby`'s test suite against a recent version of the
OpenAPI, and noticed that unfortunately, recent updates have broken it.
In particular, the new tiers feature under plan creation now has a
primitive type nested in an `anyOf`, which is something that we've never
had to handle before:

``` yaml
up_to:
  anyOf:
  - enum:
    - inf
    type: string
  - type: integer
```

`stripe-ruby` tries to run this against it:

``` ruby
plan = Stripe::Plan.create(
  ...
  tiers: [{ up_to: 100, amount: 1000 }, { up_to: "inf", amount: 2000 }]
)
```

What's supposed to happen is that "100" is coerced to an integer, but it's not
because the integer possibility is nested in an `anyOf` as seen above.

This patch addresses the problem by allowing the coercer to recurse into an
`anyOf` to try coercion. It's able to theoretically handle any number of
`anyOf` possibilities, and will dutifully try coercing each one until it finds
a coercable type (or it doesn't, which is okay).

I've run `stripe-ruby`'s test suite against a version of `stripe-mock` with
this patch in it to verify that it works.

r? @tmaxwell-stripe  (This one should be a little simpler to review than the other PR.)
cc @stripe/api-libraries

### DIFF
--- a/param/coercer/coercer.go
+++ b/param/coercer/coercer.go
@@ -50,7 +50,7 @@ func CoerceParams(schema *spec.Schema, data map[string]interface{}) error {
 						CoerceParams(subSchema.Items, itemValMap)
 					} else if subSchema.Items.Type != "" {
 						// Handles the case of an array of primitive types
-						itemValCoerced, ok := coercePrimitiveType(itemVal, subSchema.Items.Type)
+						itemValCoerced, ok := coerceSchema(itemVal, subSchema.Items)
 						if ok {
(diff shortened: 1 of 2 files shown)


tmaxwell-stripe:
LGTM

brandur-stripe:
Thanks!

PR #70, opened 2018-04-26
Validate `multipart/form-data` + error on mismatched media types

### LINKED ISSUE #69: Validate incoming media type
It'd be a nice addition to make sure that a request's incoming `Content-Type` is the same as the media type dictated by OpenAPI spec. Mostly this won't make a difference because we'll probably fall back to form encoding, but it'd be useful for vetting implementations against the file endpoints.

brandur-stripe:
This makes the changes necessary to be able to successfully validate
payloads sent as `multipart/form-data` along with the standard
`application/x-www-form-urlencoded`.

To accomplish this, I ended up splitting the parameter parser into two
separate components:

1. A new "parser" that just splits an incoming string into form
   key/value tuples. This is used to parse query strings and body form
   data. We prefer it over Go's built-in parser because it returns
   an ordered slice instead of an unordered map, which is critical for
   correctly parsing incoming arrays and maps.
2. An "assembler" that takes the parser's output and assembles
   higher-level data structures from it, including arrays and maps by
   using the "Rack-style" conventions in use by the Stripe API.

To parse `multipart/form-data`, we rely on Go's built-in multipart
parser, and then hands off its results to our new assembler. You can see
all of this at work in `param.ParseParams`.

In addition, it tweaks behavior so that if a client sends either an
empty `Content-Type` header where one was expected (i.e., usually
non-`GET`) or a `Content-Type` that doesn't match the expected media
type of the responding endpoint, `stripe-mock` will respond with an
error.

This last part is intended to be an extra safety feature to help ensure
that client implementations are correct. In particular, it should be
helpful in making sure that file API endpoints receive
`multipart/form-data` instead of the standard
`application/x-www-form-urlencoded`.

So the one big caveat is that file uploads in `stripe-ruby` and other
libraries will still not quite succeed against these changes. To get
them fully working we need to change the type of the "file create"
endpoint's `file` parameter from an object to a string. I have another
pull going into the backend to do this.

Fixes #69.

r? @tmaxwell-stripe  Hey sorry, this is another relatively large change. Mind taking a
quick pass over it? I think that things could always be better, but this
continues to take us in a more correct direction, further modularizes, and adds
more testing than we had before. Thanks!

cc @stripe/api-libraries

### DIFF
--- a/param/form/form.go
+++ b/param/form/form.go
@@ -1,0 +1,15 @@
+// Package form is a small package used to hold a few types common to the param
+// package so that we don't have import cycles.
+package form
+
+//
+// Public types
+//
+
+// FormPair is a key/value pair as extracted from a form-encoded string. For
+// example, "a=b" is the pair [a, b].
+type FormPair [2]string
+
+// FormValues is a full slice of all the key/value pairs from a form-encoded
+// string.
+type FormValues []FormPair
--- a/param/form/form_test.go
+++ b/param/form/form_test.go
@@ -1,0 +1,4 @@
+package form
+
+// Add tests here if the package is ever expanded beyond its currently trivial
+// state.
--- a/param/nestedtypeassembler/nestedtypeassembler.go
+++ b/param/nestedtypeassembler/nestedtypeassembler.go
@@ -1,0 +1,286 @@
+// Package nestedtypeassembler takes a collection of FormPair tuples and uses
+// them to construct more complex data types using "Rack-style" conventions for
+// arrays and maps with a few small Stripe-specific tweaks.
+//
+// When processing a request, data should first be parsed from the query, form,
+// or multipart form, handed off to this package for assembly, then passed on
+// the coercer to coerce string to other expected types.
+package nestedtypeassembler
+
+import (
+	"fmt"
+
+	"github.com/stripe/stripe-mock/param/form"
+)
+
+//
+// Public functions
+//
+
+func AssembleParams(form form.FormValues) (map[string]interface{}, error) {
+	params := make(map[string]interface{})
+
+	for _, pair := range form {
+		key := pair[0]
+		value := pair[1]
+
+		keyParts := parseKey(key)
+
+		if len(keyParts) == 0 {
+			continue
+		}
+
+		rawkeyPart := keyParts[0]
+		if rawkeyPart.KeyType() != keyTypeRaw {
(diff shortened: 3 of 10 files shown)


param/assembler/assembler.go
```diff
@@ -0,0 +1,286 @@
+// Package assembler takes a collection of FormPair tuples and uses them to
+// construct more complex data types using "Rack-style" conventions for arrays
+// and maps with a few small Stripe-specific tweaks.
+//
+// When processing a request, data should first be parsed from the query, form,
+// or multipart form, handed off to this package for assembly, then passed on
+// the coercer to coerce string to other expected types.
+package assembler
```
tmaxwell-stripe: The corresponding function in Rack is called `Rack::Utils.parse_nested_query`. I wonder if we could choose a name that suggests the connection?

tmaxwell-stripe:
@brandur-stripe Did you make any changes to the parameter parsing code aside from rearranging and renaming it?

param/assembler/assembler.go
brandur-stripe: Yeah. I'd kind of like to have a name that doesn't include "parse" though, because we have other modules that do the real string parsing and are more deserved of the name.

I don't mind "assembler" that much. I'm trying to think of some alternatives like "typebuilder", but don't like how that can so easily be confused with types from the programming language. Maybe "nestedformbuilder"? That's starting to get a little wordy though.

brandur-stripe:
> @brandur-stripe Did you make any changes to the parameter parsing code aside from rearranging and renaming it?

No, a few things have been moved around, but the new `parser` and `assembler` are almost entirely what was in just `parser` before. The (relatively small amount of code) in `param` is new, and may be more worthy of review.

param/assembler/assembler.go
tmaxwell-stripe: Maybe `nestedqueryassembler`? Or even just `nestedquery`? I'll leave it up to you.

tmaxwell-stripe:
LGTM

param/assembler/assembler.go
brandur-stripe: Thanks! Alright, I ended up going with `nestedtypeassembler`.

brandur-stripe:
Bringing this in. I'll cut a new release when we have an updated OpenAPI spec.

PR #73, opened 2018-05-10
Allow empty content type on `DELETE`

brandur-stripe:
We recently added checks in `stripe-mock` to make sure that the right
`Content-Type` was being sent. Unfortunately, the logic was not quite
right for deletes, where you occasionally see a `Content-Type` in rare
cases, but which usually have empty request bodies, and therefore an
empty `Content-Type` is allowed.

This patch brings in two fixes:

1. The OpenAPI spec has been tweak so as to not incorrectly include
   `expand` parameters on `DELETE` operations. This means that most
   `DELETE` operations have no defined request schema at all.
2. Special cases incoming `DELETE` operations so that they can be
   checked against a schema if they provide parameters, but are also
   allowed to specify an empty `Content-Type` as well.

Between the two improvements, all edges should be covered.

The impetus here is to resolve [1] where a `stripe-mock` upgrade is
incompatible with the current code. I've verified that with this patch
included, `stripe-python`'s test suite runs successfully.

[1] https://github.com/stripe/stripe-python/pull/426


main_test.go
```diff
@@ -83,14 +83,26 @@ func initTestSpec() {
 				Content: map[string]spec.MediaType{
 					"application/json": {
 						Schema: &spec.Schema{
-							Ref: "#/components/schemas/customer",
+							Ref: "#/components/schemas/charge",
```
brandur-stripe: This tweaks is an unrelated fix.

server.go
```diff
@@ -586,6 +546,73 @@ func parseExpansionLevel(raw []string) *ExpansionLevel {
 	return level
 }
 
+// validateAndCoerceRequest validates an incoming request against an OpenAPI
+// schema and does parameter coercion.
+//
+// Firstly, `Content-Type` is checked against the schema's media type, then
+// string-encoded parameters are coerced to expected types (where possible).
+// Finally, we validate the incoming payload against the schema.
+func validateAndCoerceRequest(
```
brandur-stripe: Most of this code is the same, but for cleanliness I've pulled it into its own function.

server_test.go
```diff
@@ -178,7 +186,7 @@ func TestStubServer_RoutesRequest(t *testing.T) {
 			&http.Request{Method: "POST",
 				URL: &url.URL{Path: "/v1/invoices/in_123/pay"}})
 		assert.NotNil(t, route)
-		assert.Equal(t, chargeDeleteMethod, route.operation)
+		assert.Equal(t, invoicePayMethod, route.operation)
```
brandur-stripe: My tweaks to the test fixtures revealed that these were previously wrong, so they've also been fixed.

brandur-stripe:
Going to self-approve this one so we can start closing out these upgrades:

LGTM.

PR #74, opened 2018-05-16
Replace data in response with data from request

brandur-stripe:
Here we try to to some additional enrichment on response data by
replacing what we found in fixtures with data that came in from the
request.

The algorithm is quite simple: we walk the decoded request object and
the generated response object, replace any scalars that we see where the
key name is the same, and recurse in when we see two keys with the same
names that are both maps.

In addition, I realized that we are no longer vetting that incorrect
keys are not being passed to stripe-mock, which appears to be a
regression from a change passed some time ago. This patch also adds back
a check on `additionalProperties: false`.

cc @tmaxwell-stripe Any thoughts on this?
cc @stripe/api-libraries

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -6,6 +6,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/stripe/stripe-mock/generator/datareplacer"
 	"github.com/stripe/stripe-mock/spec"
 )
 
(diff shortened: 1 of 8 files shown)


tmaxwell-stripe:
:+1: in general, but I think we should look at the response schema type when deciding whether or not to replace data, in addition to or instead of looking at the fixture data type.

brandur-stripe:
@tmaxwell-stripe Thanks Tim! And yes, totally agreed. With this framework in place, I'm going to add that on as the next incremental step.

brandur-stripe:
Alright, going to pull this in for now, and keep iterating.

PR #77, opened 2018-05-31
Reflect `Idempotency-Key` header back in response

brandur-stripe:
Behave like the real Stripe API by reflecting an idempotency key that
was sent into stripe-mock back in the response body.


brandur-stripe:
Released as 0.17.0.

PR #79, opened 2018-06-09
Update bundled OpenAPI and fixtures

### LINKED ISSUE #78: Customer#default_source not present in response
Hi,

I'm having this issue testing in Ruby, using stripe-mock, in which `stripe_customer.default_source` raises a `NoMethodError`. I'm afraid this is because responses from stripe-mock don't include this attribute.

brandur-stripe:
The biggest change here is that we now include fixture data for
resources with fields that were optional and had `nil` values. There was
previously a bug that prevented this.

Fixes #78.


brandur:
I also ended up adding nominal support for `format`, `maxLength`, and `x-stripeBypassValidation` here because those are all new since the last time we bumped the OpenAPI spec here apparently. Of those, `maxLength` validation will be supported, but the other two are largely ignored for now.

PR #80, opened 2018-07-06

rattrayalex-stripe:
r? @brandur-stripe || @remi-stripe


brandur-stripe:
Ah, sorry for the delay here, Alex. LGTM.

PR #83, opened 2018-07-18
Update OpenAPI

brandur-stripe:
It seems that I regressed payment intents during the last release (I
think something wasn't committed). Here we update OpenAPI again to make
sure that we have everything.


brandur-stripe:
Welp, it looks like this will be my forcing function for getting fixtures in for the new deleted objects.

brandur-stripe:
> Welp, it looks like this will be my forcing function for getting fixtures in for the new deleted objects.

Alright, they're now in!

brandur-stripe:
Released as 0.20.1.

PR #84, opened 2018-07-18
Support simultaneous HTTP and HTTPS

brandur-stripe:
I was a little hasty in adding support for HTTPS before in that when
we're developing with stripe-mock locally, we need to have a setup that
supports both HTTP *and* HTTPS because some libraries will be using one
and some libraries will be using the other.

I've added a number of command line options that allow stripe-mock to be
booted so that it's listening for both HTTP and HTTPS (and HTTP/2)
simultaneously. It's added a little bit of complication, so I've added a
number of checks for incompatible options and more documentation to the
README on how to use them all.

`12111` will remain our default port for HTTP, and `12112` will become
our default port for HTTPS. Libraries that want HTTP will lock to
`12111` by default and those that need HTTPS will lock to `12112`.

See the README for full details, but here are some basic usages with the
options:

``` sh
# listen for HTTP on port 12111
stripe-mock

# listen for HTTPS on port 12112
stripe-mock -https

# listen for HTTP on port 80
stripe-mock -port 80

# listen for HTTPS on Unix socket
stripe-mock -https -unix /tmp/stripe-mock-secure.sock

# listen for HTTP on port 80 and HTTPS on port 443
stripe-mock -http-port 80 -https-port 443
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,7 +18,8 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-const defaultPort = 12111
+const defaultPortHTTP = 12111
+const defaultPortHTTPS = 12112
 
 // verbose tracks whether the program is operating in verbose mode
 var verbose bool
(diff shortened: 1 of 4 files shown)


brandur-stripe:
r? @ob-stripe 
cc @stripe/api-libraries

main.go
```diff
@@ -103,15 +128,141 @@ func main() {
...
+
+	if o.httpsUnixSocket != "" && o.httpsPort != 0 {
+		return fmt.Errorf("Please specify only one of -https-port or -https-unix")
+	}
+
+	return nil
+}
+
+// getHTTPListener gets a listener on a port or unix socket depending on the
+// options provided. If HTTPS should not be enabled, it returns nil.
```
ob-stripe: Should this say "If HTTP should not be enabled" instead?

brandur-stripe: Oops. Yep, fixed now. Thanks!

PR #85, opened 2018-07-19
Listen for HTTP and HTTPS simultaneously in .plist

brandur-stripe:
See stripe/stripe-mock#84 for full context, but here we change
Goreleaser's .plist so that when the Homebrew tap is used, we listen on
both port 12111 and port 12112 for HTTPS.

I originally opened this at stripe/homebrew-stripe-mock#4, but forgot
that that wasn't the right place to change this setting. It needs to be
done in `goreleaser.yml` instead.

### DIFF
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -29,8 +29,10 @@ brew:
         <key>ProgramArguments</key>
         <array>
           <string>#{opt_bin}/stripe-mock</string>
-          <string>-port</string>
+          <string>-http-port</string>
           <string>12111</string>
+          <string>-https-port</string>
+          <string>12112</string>
         </array>
         <key>RunAtLoad</key>
         <true/>


brandur-stripe:
FYI @remi-stripe @ob-stripe, the next time you upgrade stripe-mock, it might be a good idea to just double check your .plist file to make sure that it includes both a port for HTTP and a port for HTTPS.

I'm pretty sure this is all that's necessary to get the right .plist (it worked for me):

``` sh
brew upgrade stripe-mock
brew services restart stripe/stripe-mock/stripe-mock
```

But afterwards you might want to:

``` sh
cat ~/Library/LaunchAgents/homebrew.mxcl.stripe-mock.plist
```

And just look in there that both `-http-port` and `-https-port` are present.

ob-stripe:
👍  Yep, worked for me.

PR #86, opened 2018-07-20
Remove support for YAML files

brandur-stripe:
Removes support for YAML files. It turns out that these don't really
work anyway, and since we're keeping JSON in the main repository these
days, it's somewhat extraneous.

r? @remi-stripe
cc @stripe-internal/api-libraries


main.go
```diff
@@ -289,43 +287,24 @@ func getTLSCertificate() (tls.Certificate, error) {
...
 
 	if fixturesPath == "" {
 		// And do the same for fixtures
 		data, err = Asset("openapi/openapi/fixtures3.json")
 	} else {
 		data, err = ioutil.ReadFile(fixturesPath)
-
-		if filepath.Ext(fixturesPath) == ".yaml" {
-			isYAML = true
-		}
```
remi-stripe: is it worth returning a friendly error if the extension is not json in case people were using yaml before?

brandur-stripe: Good call. Added an additional check in 16d5511.

PR #87, opened 2018-07-23
Add README sections for current state and future plans

brandur-stripe:
We had a report from a user that it's somewhat difficult to figure out
stripe-mock actually does today. Here we add a few new sections to the
README that describe its current state, its limitations, and what we'd
like to implement down the road.

r? @ob-stripe
cc @stripe/api-libraries


brandur-stripe:
Thank you!

PR #88, opened 2018-07-25
Whitelist RPC-style actions `/approve` and `/decline`

brandur-stripe:
stripe-mock currently needs a whitelist of RPC-style names so that it
can differentiate between an RPC endpoint and a subcollection when
deciding whether to include the primary ID from a URL.

Here we add two new ones for `/approve` and `/decline`.

r? @zanker-stripe
cc @remi-stripe


zanker-stripe:
LGTM

brandur-stripe:
Thanks!

PR #89, opened 2018-07-25
Decide between returning a deleted versus non-deleted from HTTP method

brandur-stripe:
Implements a simple heuristic so that stripe-mock will pick an `anyOf`
branch based off whether the incoming HTTP request method is a `DELETE`
or or not. If it is `DELETE`, we try to pick the deleted form of a
resource, and vice versa.

We're doing this because we'd occasionally run into a place where
stripe-mock was picking a branch based off its naive "pick the first
branch" logic when it was quite obvious that another branch would
overwhelmingly have been preferred. For example, returning a deleted
customer from `GET /v1/customers/:id` because the deleted customer
appeared first in the `anyOf` list.

The heuristic isn't perfect, but it will usually do the right thing.

r? @ob-stripe
cc @stripe/api-libraries

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -42,6 +42,14 @@ type GenerateParams struct {
 	// they'd otherwise be if they'd been generated from fixtures alone..
 	RequestData map[string]interface{}
 
+	// RequestMethod is the HTTP method of the URL being requested which we're
+	// generating data for. It's used to decide between returning a deleted and
+	// non-deleted schema in some cases.
+	//
+	// The value of this field is expected to stay stable across all levels of
+	// recursion.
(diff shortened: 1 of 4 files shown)


brandur-stripe:
Thanks!

brandur-stripe:
Released as v0.22.0.

brandur-stripe:
(And @remi-stripe, would you mind checking to see if this helps your testing situation?)

remi-stripe:
@brandur-stripe I just upgraded and the test suite on stripe-php now passes again. Thank you for the quick fix!!!!

PR #90, opened 2018-07-25
Activate golint for stripe-mock

brandur-stripe:
Adds golint and fixes the places that it didn't like. This is a no-op
functionally speaking -- the changes are largely around naming and
comments.


brandur-stripe:
These are non-functional changes and it's a large-ish diff, so going to self-approve this one.

PR #92, opened 2018-07-30
Update OpenAPI (to get scheduled query runs)

brandur-stripe:
Updates the bundled OpenAPI specification, most notably getting us
support for scheduled query runs.


brandur-stripe:
Released as 0.24.0.

PR #93, opened 2018-07-30
Don't replace data on non-`POST` endpoints

brandur-stripe:
stripe-mock has a feature where it reflects data from input parameters
back into the response so that when creating an updating objects, it
produces responses that look more realistic with some of the actual
input values that were sent in.

Unfortunately, we found that this was working a little _too_ well and
being applied to other operations like listing as well. So for example,
if we request a list with a filter of `object=source`, we'd see a
response where the response list object's `object` field was set to
`source` instead of `list` like it should be.

Here we make sure that data replacement only occurs on `POST`
operations and add some tests to verify this behavior.

r? @remi-stripe
cc @dpetrovics-stripe

### DIFF
--- a/generator_test.go
+++ b/generator_test.go
@@ -264,6 +264,38 @@ func TestGenerateResponseData(t *testing.T) {
 			data.(map[string]interface{})["customer"])
 	}
 
+	// data replacement on `POST`
+	{
+		generator := DataGenerator{testSpec.Components.Schemas, &testFixtures}
+		data, err := generator.Generate(&GenerateParams{
+			RequestData: map[string]interface{}{
+				"customer": "cus_9999",
+			},
+			RequestMethod: http.MethodPost,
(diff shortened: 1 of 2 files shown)


generator_test.go
```diff
@@ -264,6 +264,38 @@ func TestGenerateResponseData(t *testing.T) {
 			data.(map[string]interface{})["customer"])
 	}
 
+	// data replacement on `POST`
+	{
+		generator := DataGenerator{testSpec.Components.Schemas, &testFixtures}
+		data, err := generator.Generate(&GenerateParams{
+			RequestData: map[string]interface{}{
+				"customer": "ch_9999",
```
remi-stripe: really minor but that looks like a charge id

brandur-stripe: Oops. Yep, thanks for that! My fingers just typed the wrong combination of letters. Fixed.

brandur-stripe:
Thanks Remi!

Released as 0.24.1.

PR #98, opened 2018-08-07
Fix test to match naming + less aggressive parameter coercion

brandur-stripe:
Remi saw on #97 that I'd renamed a field to `bool_or_int_key` intending
to change a test, but I'd accidentally forgotten to make that subsequent
change, which was to create a higher fidelity test by testing two
primitive type schemas within the same `anyOf`.

Here I made that change, but noticed when I did it that the tests
started to fail. It turned out this is because we were coercing
primitive types overly aggressively. If for example a boolean schema
received an integer like `123`, `strconv.ParseBool` would fail with an
error, but we'd just default to the value `false`.

In this test, it had the effect of the `anyOf` taking the first boolean
branch, even though the value is much better suited to be coerced to an
integer.

Here we relax primitive type coercion. If a value can be parsed for that
type, it is, but if it's not, we default to the value that was given to
us. This seems like a better fit for the coercion behavior that we
actually want.

r? @remi-stripe

### DIFF
--- a/param/coercer/coercer.go
+++ b/param/coercer/coercer.go
@@ -105,21 +105,21 @@ func coercePrimitiveType(val interface{}, primitiveType string) (interface{}, bo
 	case primitiveType == booleanType:
 		valBool, err := strconv.ParseBool(valStr)
 		if err != nil {
-			valBool = false
+			return nil, false
 		}
 		return valBool, true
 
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Thanks Remi!

brandur-stripe:
Released as 0.25.2.

PR #99, opened 2018-08-20
Update OpenAPI

brandur-stripe:
Specifically pulls in usage record summaries (along with any other
changes that have occurred recently).


brandur-stripe:
Released as 0.26.0.

PR #100, opened 2018-08-23
Parse form values from the query string even for non-`GET` requests

brandur-stripe:
The Stripe API will still take values from the query string even if for
non-`GET` requests, and even if a request body was sent as well. From
the API's perspective internally, these all look the same and come in as
one big bucket of parameters. This behavior is inherited from Rack,
which has been doing this forever.

Here we try to more closely match the API's true behavior by allowing
query string values even for non-`GET` requests.

r? @remi-stripe Hopefully this will fix the problem you're seeing.
cc @stripe/api-libraries


brandur-stripe:
> LGTM, I did not expect such a fast ship!

Anything to get that .NET work unblocked ;)

Thanks!

PR #105, opened 2018-08-27
Update bundled OpenAPI

brandur-stripe:
Most notably, this brings in request body definitions even for endpoints
that don't take parameters like most `DELETE`s.


brandur-stripe:
Released as 0.29.0.

PR #112, opened 2018-09-14
Add x-stripeResource to supported params

tomer-stripe:
This ignores the `x-stripeResource` field we added to the openapi spec, we don't need that for the mock server.

I ran `pushd openapi/ && git pull origin master && popd` and `go generate` which resulted in the changes to `bindata.go` and the `openapi` submodule. Let me know if I should undo those.


brandur-stripe:
LGTM!

tmaxwell-stripe:
LGTM

PR #121, opened 2018-10-19
Add new "action" methods to ID replacement whitelist

brandur-stripe:
Okay, we have to get this whitelist eliminated, but here we add our new
"action-style" endpoints for now.

r? @remi-stripe


brandur:
Thanks Remi!

brandur-stripe:
Released as v0.34.1.

PR #123, opened 2018-10-26

remi-stripe:
r? @stripe/dev-platform 

This adds support for the Person resource so that we can release the libraries.


ob-stripe:
Released as 0.35.0.

PR #138, opened 2019-01-30
Add /release to the list of actions that can return a resource

remi-stripe:
An upcoming feature called Subscription Schedules have the /release action. stripe-mock has a list of those actions so that we know to replace the resource id with a fixture one instead.

r? @brandur-stripe 
cc @stripe/api-libraries


ob-stripe:
Released as v0.43.2.

PR #141, opened 2019-02-18
Handle `anyOf` schema containing array in params coercer

mickjermsurawong-stripe:
- Currently stripe-mock doesn't expect array in `anyOf` schema. When it finds non-primitive sub-schema of `anyOf` it recurses using the main `CoerceParams` method. The method is used for top-level coercion, and assumes schema having properties. Because `Items` sub-schema for the array does not have properties, and it will get dropped. The output will remain as integer-indexed map value. For example, invoice update operation in OpenAPI has param for `custom_fields` as follows:
```
custom_fields:
  anyOf:
  - items:
      properties:
        // ...
      type: object
    type: array
  - enum:
    - ''
    type: string
```
and the resulting stripe-mock error is:
```
Request: POST /v1/invoices/in_123
Request data = map[custom_fields:map[0:map[value:val1 name:foo1] 1:map[name:foo2 value:val2]]]
Request validation error: validator 0xc000a182d0 failed: object property 'custom_fields' validation failed: could not validate against any of the constraints
```
- Verified that after this fix the output in stripe-mock is now
```
Request data = map[custom_fields:[map[name:foo1 value:val1] map[name:foo2 value:val2]]]
```
- The implementation introduces `coerceSubSchema` with different recurse behavior based on whether it is `object` schema. 
- The motivation here is when I'm testing for polymorphic params for java sdk-autogen and came across this issue.

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/param/coercer/coercer.go
+++ b/param/coercer/coercer.go
@@ -19,55 +19,35 @@ func CoerceParams(schema *spec.Schema, data map[string]interface{}) error {
 		if !ok {
 			continue
 		}
-
-		valMap, ok := val.(map[string]interface{})
-		if ok {
-			CoerceParams(subSchema, valMap)
-
-			if subSchema.Type == arrayType {
-				valSlice, err := parseIntegerIndexedMap(valMap)
-				if err != nil {
-					return err
-				}
-
-				if valSlice != nil {
-					data[key] = valSlice
-					val = valSlice
-				}
-			}
-
-			// May fall through to the next segment where we iterate an array
-			// and coerce it
+		val, ok, err := coerceSubSchema(val, subSchema)
+		if err != nil {
+			return err
 		}
-
-		valArr, ok := val.([]interface{})
 		if ok {
-			if subSchema.Items != nil {
-				for i, itemVal := range valArr {
-					itemValMap, ok := itemVal.(map[string]interface{})
-					if ok {
-						// Handles the case of an array of generic objects
-						CoerceParams(subSchema.Items, itemValMap)
-					} else if subSchema.Items.Type != "" {
-						// Handles the case of an array of primitive types
-						itemValCoerced, ok := coerceSchema(itemVal, subSchema.Items)
-						if ok {
-							valArr[i] = itemValCoerced
-						}
-					}
-				}
-			}
-
-			continue
-		}
-
-		valCoerced, ok := coerceSchema(val, subSchema)
-		if ok {
-			data[key] = valCoerced
+			data[key] = val
 		}
 	}
 
 	return nil
+}
+
+// coerceSubSchema coerces sub-param value according to an arbitrary JSON sub-schema.
+// It is sub-schema because it can be array, primitive, or object,  unlike the main `CoerceParams`
+// only expecting object type with properties. This is also used in coercing each sub-schema
+// of anyOf or array.
+func coerceSubSchema(val interface{}, subSchema *spec.Schema) (interface{}, bool, error) {
+	if len(subSchema.Properties) == 0 {
+		// anyOf, array, and primitive schemas
+		// additionalProperties without properties will not be coerced
+		return coerceNonObjectSchema(val, subSchema)
+	}
+	// schema with properties with `object` type
+	valMap, ok := val.(map[string]interface{})
+	var err error
+	if ok {
+		err = CoerceParams(subSchema, valMap)
+	}
+	return valMap, ok, err
 }
 
 //
@@ -80,6 +60,7 @@ const (
 	booleanType = "boolean"
 	integerType = "integer"
 	numberType  = "number"
+	stringType  = "string"
 	objectType  = "object"
 )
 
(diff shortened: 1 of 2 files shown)


param/coercer/coercer.go
```diff
@@ -19,57 +19,35 @@ func CoerceParams(schema *spec.Schema, data map[string]interface{}) error {
...
+		// additionalProperties without properties will not be coerced
+		return coerceNonObjectSchema(val, subSchema)
+	}
+	// schema with properties with `object` type
+	valMap, ok := val.(map[string]interface{})
+	var err error
+	if ok {
+		err = CoerceParams(subSchema, valMap)
+	}
+	return valMap, ok, err
```
mickjermsurawong-stripe: The controlling logic into different recurse behavior of a given sub-schema

param/coercer/coercer.go
```diff
@@ -19,57 +19,35 @@ func CoerceParams(schema *spec.Schema, data map[string]interface{}) error {
 		if !ok {
 			continue
 		}
-
-		valMap, ok := val.(map[string]interface{})
-		if ok {
-			CoerceParams(subSchema, valMap)
-
-			if subSchema.Type == arrayType {
-				valSlice, err := parseIntegerIndexedMap(valMap)
```
mickjermsurawong-stripe: Converting integer-indexed map to array is handle within `coerceNonObjectSchema` together with other array-related logic

param/coercer/coercer.go
```diff
@@ -122,40 +101,78 @@ func coercePrimitiveType(val interface{}, primitiveType string) (interface{}, bo
 			return nil, false
 		}
 		return valFloat, true
+
+	case primitiveType == stringType:
```
mickjermsurawong-stripe: Added check for string primitive because in `anyOf` the first one is chosen. For example when given ["123", "456"], it should not be coerced to [123, 456] if string type is explicitly given. 
See test ``anyOf` with array of different primitive`

param/coercer/coercer.go
```diff
@@ -122,40 +101,71 @@ func coercePrimitiveType(val interface{}, primitiveType string) (interface{}, bo
...
-// coerceSchema tries to coerce a schema containing a primitive type from the
-// given generic interface{} value.
+// coerceNonObjectSchema tries to coerce a non-object schema given generic interface{} value.
 //
 // It's similar to coercePrimitiveType above (and indeed calls into it), but
-// also handles the case of an anyOf schema that supports a number of different
-// primitve types.
-func coerceSchema(val interface{}, schema *spec.Schema) (interface{}, bool) {
+// also handles array and anyOf schema (supporting a number of different primitive types)
+func coerceNonObjectSchema(val interface{}, schema *spec.Schema) (interface{}, bool, error) {
```
mickjermsurawong-stripe: `anyOf` and `array` block below will recurse its sub-schema with `coerceSubSchema`

param/coercer/coercer.go
```diff
@@ -19,57 +19,38 @@ func CoerceParams(schema *spec.Schema, data map[string]interface{}) error {
...
 		if ok {
-			data[key] = valCoerced
+			data[key] = coercedVal
 		}
 	}
 
 	return nil
 }
 
+// coerceSubSchema coerces sub-param value according to an arbitrary JSON sub-schema.
```
brandur-stripe: Sorry these are all wording nits, but could we say "coerces a sub-param value"?

param/coercer/coercer.go
```diff
@@ -19,57 +19,38 @@ func CoerceParams(schema *spec.Schema, data map[string]interface{}) error {
...
+// It is sub-schema because it can be array, primitive, or object,  unlike the main `CoerceParams`
```
brandur-stripe: Let's say:

> It is named with "sub-schema" because ...

param/coercer/coercer.go
```diff
@@ -19,57 +19,38 @@ func CoerceParams(schema *spec.Schema, data map[string]interface{}) error {
...
+// only expecting object type with properties. This is also used in coercing each sub-schema
+// of anyOf or array.
+func coerceSubSchema(val interface{}, subSchema *spec.Schema) (interface{}, bool, error) {
+	if len(subSchema.Properties) == 0 {
+		// Non-object schemas are anyOf, array, and primitive schemas.
+		// Implicitly treats actual object schema with empty properties as non-object.
+		return coerceNonObjectSchema(val, subSchema)
+	}
+	// `object` schema with properties
```
brandur-stripe: Can you put a blank line above this comment?

param/coercer/coercer.go
```diff
@@ -80,6 +61,7 @@ const (
 	booleanType = "boolean"
 	integerType = "integer"
 	numberType  = "number"
+	stringType  = "string"
```
brandur-stripe: Can you keep this ordered alphabetically so that `stringType` appears below `objectType`?

param/coercer/coercer.go
brandur-stripe: There's also two spaces before "unlike" — can you trim them down to just one?

mickjermsurawong-stripe:
ptal @brandur-stripe 
Thank you for the review! I addressed all the suggestions on style.
Working with Go is actually pleasant!

brandur-stripe:
LGTM! Thanks (I'll take release).

> Working with Go is actually pleasant!

Always good to have another convert :) Amazing how quickly you picked it up!

brandur-stripe:
Released as v0.46.0.

PR #142, opened 2019-02-28
Fix handling anyOf enum and number

mickjermsurawong-stripe:
- This PR handles anyOf schema of integer and enum. The problem is any string input now satisfies Enum string-ness coercion; we don't actually check the specific type of enum (that is done later).
When given numeric string, it naively satisfies the enum as well, and the validator later complains that the numeric string isn't the integer type. The problematic schema is:
```
anyOf:
- enum:
  - inf
  maxLength: 5000
  type: string
- type: integer
```
- This isn't a problem before trying to handle anyOf Array and code structuring in this PR https://github.com/stripe/stripe-mock/pull/141/files because String isn't considered a primitive and always un-coerced, and thus numeric string will be always be considered as coercion candidate. 
- Looking at that PR again, string coercion helps to make `coerceSubSchema` return type consistent. 
Parsing sub-schema of AnyOf or Items can use "coerced" flag output consistently. 
- I can do code restructuring to consider String coerced, but I think it will be more complex than this additional assumption in this PR.
- This code will fail if we ever allow enum as numeric "123", "0.5" or if as "true"/"false"
cc @stripe/api-libraries

### DIFF
--- a/param/coercer/coercer_test.go
+++ b/param/coercer/coercer_test.go
@@ -126,6 +126,40 @@ func TestCoerceParams_AnyOfCoercion(t *testing.T) {
 		assert.Equal(t, nil, sliceVal[1])
 		assert.Equal(t, "456", sliceVal[2])
 	}
+
+	// `anyOf` Enum and numeric string
+	{
+		schema := &spec.Schema{Properties: map[string]*spec.Schema{
+			"enum_or_int_key": {
+				AnyOf: []*spec.Schema{
+					{Enum: []interface{}{
+						"enum1",
+						"enum2",
+					}, Type: stringType,
+					},
+					{Type: integerType},
+				},
+			},
+		}}
+
+		// numeric string shouldn't be considered as enum string
+		// and still be coerced to number
+		data := map[string]interface{}{
+			"enum_or_int_key": "123",
+		}
+		err := CoerceParams(schema, data)
+		assert.NoError(t, err)
+		assert.Equal(t, 123, data["enum_or_int_key"])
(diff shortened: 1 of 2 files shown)


param/coercer/coercer_test.go
```diff
@@ -126,6 +126,40 @@ func TestCoerceParams_AnyOfCoercion(t *testing.T) {
...
+	// `anyOf` Enum and numeric string
+	{
+		schema := &spec.Schema{Properties: map[string]*spec.Schema{
+			"enum_or_int_key": {
+				AnyOf: []*spec.Schema{
+					{Enum: []interface{}{
+						"enum1",
+						"enum2",
+					}, Type: stringType,
+					},
```
remi-stripe: minor but is this indentation expected?

mickjermsurawong-stripe:
thanks @remi-stripe fixed the indentation!
self-approving
LGTM

mickjermsurawong-stripe:
Besides `stripe-java`, also verified that `stripe-go` (on master) that first failed on also passed with this latest stripe-mock now.
```
                        Error:          Expected nil, but got: &stripe.Error{ChargeID:"", Code:"", Err:(*stripe.InvalidRequestError)(0xc42000e0c8), HTTPStatusCode:400, Msg:"Request validation error: validator 0xc000912e70 failed: object property 'tiers' validation failed: object property 'up_to' validation failed: could not validate against any of the constraints", Param:"", RequestID:"req_123", Type:"invalid_request_error"}
```

PR #145, opened 2019-03-05
Add new API operations with custom words

remi-stripe:
This adds new endpoints name that unblock test suites in our client libraries for new resources `PaymentIntent` and `PaymentMethod`.

r? @brandur-stripe 
cc @stripe/api-libraries 

We likely need a better way moving forward as we're committing to more and more "verb" actions though.


brandur-stripe:
LGTM.

> We likely need a better way moving forward as we're committing to more and more "verb" actions though.

Yep, totally :$ There's probably some way to extract this out of the API spec now, and if not, it'd be worth adding a vendor extension to get this hack torn out.

PR #146, opened 2019-03-06
[WIP] stripe-mock using params reference

mickjermsurawong-stripe:
- This is experimental PR trying to assess the scope of stripe-mock change if we need to consume params with `$ref: #components..` to the components schema.
- It doesn't seem like too much a change, so I'm not sure if I'm missing something. My understanding is the request schema currently doesn't support param refs, hence my change here. However, the validator seems to be working well with the param refs.
- I tested locally consuming spec with param refs in https://git.corp.stripe.com/stripe-internal/pay-server/pull/134851/files and testing against java test-suite works fine. 
- If we want to make stripe-mock support param refs, I will write more tests and perhaps extract the logic to a more appropriate place. I would appreciate your suggestion at this stage too.
- Motivation is that spec file at least `dashboard.json` is hitting limit currently 4.9MB. The extracted param with refs in the mentioned PR can give us 4.2MB

r? @brandur-stripe 
@stripe/api-libraries

### DIFF
--- a/server.go
+++ b/server.go
@@ -263,16 +263,14 @@ func (s *StubServer) initializeRouter() error {
 			// (because it became ossified in Stripe's server implementation).
 			if verb == "get" {
 				requestSchema = spec.BuildQuerySchema(operation)
-
 				var err error
 				requestValidator, err = spec.GetValidatorForOpenAPI3Schema(
-					requestSchema, nil)
+					requestSchema, componentsForValidation)
 				if err != nil {
 					return err
 				}
 			} else {
 				requestMediaType, requestSchema = getRequestBodySchema(operation)
-
 				if requestSchema != nil {
 					var err error
 					requestValidator, err = spec.GetValidatorForOpenAPI3Schema(
@@ -297,6 +295,10 @@ func (s *StubServer) initializeRouter() error {
 					hasPrimaryID = true
 					break
 				}
+			}
+
+			if requestSchema != nil {
+				requestSchema = replaceRef(requestSchema, s.spec.Components.Schemas)
 			}
 
 			route := stubServerRoute{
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -335,6 +337,35 @@ func (s *StubServer) initializeRouter() error {
 	return nil
 }
 
+func replaceRef(requestSchema *spec.Schema, componentSchemas map[string]*spec.Schema) *spec.Schema {
```
brandur-stripe: I'd probably call this something like `dereferenceRef` instead.

server.go
```diff
@@ -335,6 +337,35 @@ func (s *StubServer) initializeRouter() error {
...
+	}
+
+	if len(requestSchema.Properties) != 0 {
+		for prop, val := range requestSchema.Properties {
+			requestSchema.Properties[prop] = replaceRef(val, componentSchemas)
+		}
+		return requestSchema
+	}
+
+	if len(requestSchema.AnyOf) != 0 {
```
brandur-stripe: This may be premature optimization, but it might be useful to provide a helper for `Schema` that allows a caller to iterate over the sub-schemas that will be found in it (just in case support for `oneOf` or something else is added in the future). Perhaps it could take a callback and allow the subschema to be modified by returning a non-`nil` value:

``` go
func replaceRef(requestSchema *spec.Schema, componentSchemas map[string]*spec.Schema) *spec.Schema {
    requestSchema.IterateSubSchemas(func(s *Schema) {
        return replaceRef(s, componentSchemas)
    })
    return requestSchema
}
```

Just an idea though! It's probably not necessary to do something like this right away.

mickjermsurawong-stripe:
Thank you @brandur-stripe for your review!
From the discussion, however, we are going to simply ignore the size limit on OpenAPI specs instead.
I'm closing this PR first. If we do revisit the problem, I can do a follow-up on this.

PR #147, opened 2019-03-15
Add new option `-strict-version-check`

brandur-stripe:
Adds a new option called `-strict-version-check` that causes stripe-mock
to check that any value passed in an incoming request's `Stripe-Version`
header matches the version specified in the bundled OpenAPI
specification.

This is intended to act as an optionally strengthened expectation which
protects against an unintended drift between a bundled stripe-mock /
OpenAPI and the version being used from an API library. We'll likely
want to enable the option for static libraries like stripe-go and
stripe-dotnet.

r? @ob-stripe

### DIFF
--- a/spec/spec.go
+++ b/spec/spec.go
@@ -49,6 +49,14 @@ type Fixtures struct {
 // HTTPVerb is a type for an HTTP verb like GET, POST, etc.
 type HTTPVerb string
 
+// Info is the `info` portion of an OpenAPI specification that contains meta
+// information about it.
+type Info struct {
+	// Version is the Stripe API version represented in the specification. It
+	// takes a date-based form like `2019-02-19`.
+	Version string `json:"version"`
+}
+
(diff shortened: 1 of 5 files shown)


spec/spec.go
```diff
@@ -49,6 +49,14 @@ type Fixtures struct {
 // HTTPVerb is a type for an HTTP verb like GET, POST, etc.
 type HTTPVerb string
 
+// Info is the `info` portion of an OpenAPI specification that contains meta
+// information about it.
+type Info struct {
```
brandur-stripe: I wanted to call this `SpecInfo` because it's more description, but Golint tells me that the import of `spec.SpecInfo` "stutters" in that there's a repetition of "spec". It's not that big of a deal, so I just called it `Info` instead.

brandur-stripe:
> Thanks for the quick implementation.

Haha, and thanks for the quick review!

brandur-stripe:
Released as 0.48.0.

PR #153, opened 2019-03-22
Generates new random primary IDs from "create" endpoints

### LINKED ISSUE #152: returned `id`s are the same for every request
Since the returned `id`s are just taken from a fixture, the same `id` is returned every time.  This causes integrity errors when testing the system I am working with

brandur-stripe:
Here we extend the ID replacement infrastructure to generate new IDs
from "create" endpoints. For example:

    $  curl -X POST -H "Content-Type: application/x-www-form-urlencoded" -H "Authorization: Bearer sk_test_123" http://localhost:2345/v1/charges | jq '.id'
    "ch_CNT2UVXcbLq0Jt5"

    $  curl -X POST -H "Content-Type: application/x-www-form-urlencoded" -H "Authorization: Bearer sk_test_123" http://localhost:2345/v1/charges | jq '.id'
    "ch_CNT2bb9wsZpgsxT"

With the idea being that much like you get with the Stripe API, new
objects have unique IDs every time.

I ported the same algorithm we use for token generation internally so
that IDs are made up of a prefix (determined from the source fixture), a
time portion, and a random portion. This means that they order in a
mostly ascending way over time (like the Stripe API this is not
guaranteed but is often true), which is always kind of nice.

Fixes #152.

r? @ob-stripe

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -2,10 +2,12 @@ package main
 
 import (
 	"fmt"
+	"math/rand"
 	"net/http"
 	"reflect"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/stripe/stripe-mock/generator/datareplacer"
 	"github.com/stripe/stripe-mock/spec"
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Thanks @ob-stripe!

> I'm fairly sure I know the answer, but just to confirm: this isn't going to affect list endpoints, correct? Even though these don't have IDs in the path, I assume the logic for those is to return a list object with a single resource fixture in the `data` attribute.

Excellent question.

It doesn't kick in for list endpoints, but it's a little hacky — the change depends on (1) no primary ID, and (2) there is an `id` field in the top-level of the response. List endpoints don't have an `id` at the top-level, so it never activates.

For now, I've added a new comment that explains this and mentions lists specifically. Longer term, it might be a good idea to add a specific annotation in the OpenAPI that indicates the "type" of every endpoint like list, create, update, delete, rpc, etc. That would allow us to make this less hacky, and also unwind a few other hacks elsewhere in the code (e.g. we currently special case list every RPC endpoint because it's hard to different these from list endpoints under a top-level resource).

PR #162, opened 2019-05-03

remi-stripe:
Add support for `replacement_for` on Issuing Card so that we can test expanding it in stripe-dotnet. I have confirmed it works locally.

r? @mickjermsurawong-stripe 
cc @stripe/dev-platform


mickjermsurawong-stripe:
LGTM

PR #163, opened 2019-05-08
Have the homebrew package commit author be 'stripe-ci'

tomer-stripe:
This has the homebrew package author be @stripe-ci instead of @goreleaserbot


brandur-stripe:
Oh man, so much better! Thanks for digging into this.

LGTM.

PR #173, opened 2019-07-22
Deduplicate slashes in incoming paths

### LINKED ISSUE #170: Customer and Subscription POST now return a `data` array in the JSON, which can't be de-serializaed to proper objects
We updated to latest stripe-mock/stripe-dontnet releases (0.60/27.10). Now both Customer and Subscription POST responses return an array of objects (the `data` field), whereas they used to return the newly added object. This appears not to be coherent with Stripe API, which still returns the object, and not an array. A quick check at the OpenAPI specification seems to confirm that a single object is still expected.

This is the `StripeResponse.Content` content following a `Stripe.CustomerService.Create()` call. The `Stripe.Customer` instance returned by this call is a non-initialized object (e.g. `Customer.Id` is null, like all other fields.)

```
"{\"data\":[{\"account_balance\":0,\"address\":null,\"balance\":0,\"created\":1234567890,\"currency\":\"usd\",\"default_source\":null,\"delinquent\":false,\"description\":null,\"discount\":null,\"email\":null,\"id\":\"cus_FJaYLWmiE8rxFC\",\"invoice_prefix\":\"180BC08\",\"invoice_settings\":{\"custom_fields\":null,\"default_payment_method\":null,\"footer\":null},\"livemode\":false,\"metadata\":{},\"name\":null,\"object\":\"customer\",\"phone\":null,\"preferred_locales\":[],\"shipping\":null,\"sources\":{\"data\":[{\"created\":1234567890,\"customer\":null,\"fingerprint\":\"OUv5A7ge4czHxnfR\",\"id\":\"aliacc_1EphH5IUzFSXAEcAfYCgLj9h\",\"livemode\":false,\"metadata\":{},\"object\":\"alipay_account\",\"payment_amount\":1000,\"payment_currency\":\"usd\",\"reusable\":false,\"used\":false,\"username\":\"test@example.com\"}],\"has_more\":false,\"object\":\"list\",\"url\":\"/v1/customers/cus_FJaYLWmiE8rxFC/sources\"},\"subscriptions\":{\"data\":[{\"application_fee_percent\":null,\"billing\":\"charge_automatically\",\"billing_cycle_anchor\":1234567890,\"billing_thresholds\":null,\"cancel_at_period_end\":false,\"canceled_at\":1234567890,\"collection_method\":\"charge_automatically\",\"created\":1234567890,\"current_period_end\":1234567890,\"current_period_start\":1234567890,\"customer\":\"cus_FJaYLWmiE8rxFC\",\"days_until_due\":null,\"default_payment_method\":null,\"default_source\":null,\"default_tax_rates\":[],\"discount\":null,\"ended_at\":1234567890,\"id\":\"sub_FKLz5hl0zyaPGd\",\"items\":{\"data\":[{\"billing_thresholds\":null,\"created\":1561579163,\"id\":\"si_FKLzxZRXl8f8qr\",\"metadata\":{},\"object\":\"subscription_item\",\"plan\":{\"active\":true,\"aggregate_usage\":null,\"amount\":2000,\"billing_scheme\":\"per_unit\",\"created\":1561402737,\"currency\":\"usd\",\"id\":\"gold\",\"interval\":\"month\",\"interval_count\":1,\"livemode\":false,\"metadata\":{},\"nickname\":null,\"object\":\"plan\",\"product\":\"prod_FJaYU37yVcOAWH\",\"tiers\":null,\"tiers_mode\":null,\"transform_usage\":null,\"trial_period_days\":null,\"usage_type\":\"licensed\"},\"quantity\":1,\"subscription\":\"sub_FKLzukkMx7NSmO\"}],\"has_more\":false,\"object\":\"list\",\"url\":\"/v1/subscription_items?subscription=sub_FKLz5hl0zyaPGd\"},\"latest_invoice\":null,\"livemode\":false,\"metadata\":{},\"object\":\"subscription\",\"plan\":{\"active\":true,\"aggregate_usage\":null,\"amount\":2000,\"billing_scheme\":\"per_unit\",\"created\":1561402737,\"currency\":\"usd\",\"id\":\"gold\",\"interval\":\"month\",\"interval_count\":1,\"livemode\":false,\"metadata\":{},\"nickname\":null,\"object\":\"plan\",\"product\":\"prod_FJaYU37yVcOAWH\",\"tiers\":null,\"tiers_mode\":null,\"transform_usage\":null,\"trial_period_days\":null,\"usage_type\":\"licensed\"},\"quantity\":1,\"start\":1234567890,\"start_date\":1234567890,\"status\":\"active\",\"tax_percent\":null,\"trial_end\":1234567890,\"trial_start\":1234567890}],\"has_more\":false,\"object\":\"list\",\"url\":\"/v1/customers/cus_FJaYLWmiE8rxFC/subscriptions\"},\"tax_exempt\":\"none\",\"tax_ids\":{\"data\":[{\"country\":\"DE\",\"created\":1234567890,\"customer\":\"cus_FJaYLWmiE8rxFC\",\"id\":\"txi_123456789\",\"livemode\":false,\"object\":\"tax_id\",\"type\":\"eu_vat\",\"value\":\"DE123456789\",\"verification\":{\"status\":\"pending\",\"verified_address\":null,\"verified_name\":null}}],\"has_more\":false,\"object\":\"list\",\"url\":\"/v1/customers/cus_FJaYLWmiE8rxFC/tax_ids\"},\"tax_info\":null,\"tax_info_verification\":null}],\"has_more\":false,\"object\":\"list\",\"url\":\"/v1/customers\"}"
```

Not sure if this is the intended behaviour, and then I am obviously missing something in my upgrade path.

brandur:
Occasionally an API library or other client might do some path appending
that's slightly off and end up with an extra slash in their target.
`//v1/charges` for example.

While the Stripe API will happily accept this, stripe-mock currently
won't because Go intercepts the doubled slashes fairly high up in the
stack and responds with a 301 to a normalized URL (e.g. `/v1/charges`).

Go's default behavior is probably desirable most of the time, but given
that the Stripe API takes doubled slashes, in this case it makes more
sense to accept them as well. This patch introduces a wrapping handler
that deduplicates slashes on incoming requests, thereby allowing a
request with doubled slashes to succeed normally.

Fixes #170.

r? @ob-stripe Mind taking a look at this one? Thanks!
cc @stripe/api-libraries

### DIFF
--- a/server.go
+++ b/server.go
@@ -21,6 +21,25 @@ import (
 // Public types
 //
 
+// DoubleSlashFixHandler is a specialized handler that wraps an HTTP mux and
+// deduplicates any doubled slashes that are included in an incoming path. So
+// `//v1/charges` would become `/v1/charges`. This works around the standard Go
+// behavior, which is to redirect the request with a 301 before it reaches the
+// underlying handler.
+//
+// The reason we deduplicate is that in some API libraries occasionally
+// generate paths with double slashes, and the real Stripe API responds to
+// these requests normally, so stripe-mock emulates that behavior.
+type DoubleSlashFixHandler struct {
+	Mux http.Handler
+}
+
+// ServeHTTP serves an HTTP request.
+func (h *DoubleSlashFixHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	r.URL.Path = strings.Replace(r.URL.Path, "//", "/", -1)
+	h.Mux.ServeHTTP(w, r)
+}
+
 // ExpansionLevel represents expansions on a single "level" of resource. It may
 // have subexpansions that are meant to take effect on resources that are
 // nested below it (on other levels).
--- a/main.go
+++ b/main.go
@@ -88,7 +88,12 @@ func main() {
 		abort(fmt.Sprintf("Error initializing router: %v\n", err))
 	}
 
-	http.HandleFunc("/", stub.HandleRequest)
+	httpMux := http.NewServeMux()
+	httpMux.HandleFunc("/", stub.HandleRequest)
+
+	// Deduplicates doubled slashes in paths. e.g. `//v1/charges` becomes
+	// `/v1/charges`.
+	handler := &DoubleSlashFixHandler{httpMux}
 
 	httpListener, err := options.getHTTPListener()
 	if err != nil {
@@ -98,7 +103,9 @@ func main() {
 	// Only start HTTP if requested (it will activate by default with no arguments, but it won't start if
 	// HTTPS is explicitly requested and HTTP is not).
 	if httpListener != nil {
-		server := http.Server{}
+		server := http.Server{
+			Handler: handler,
+		}
 
 		// Listen in a new Goroutine that so we can start a simultaneous HTTPS
 		// listener if necessary.
@@ -138,7 +145,10 @@ func main() {
 			NextProtos: []string{"h2"},
 		}
 
-		server := http.Server{TLSConfig: tlsConfig}
+		server := http.Server{
+			Handler:   handler,
+			TLSConfig: tlsConfig,
+		}
 		tlsListener := tls.NewListener(httpsListener, tlsConfig)
 
 		go func() {
--- a/server_test.go
+++ b/server_test.go
@@ -19,6 +19,62 @@ import (
 // Tests
 //
 
+func TestDoubleSlashFixHandler(t *testing.T) {
+	var lastPath string
+
+	httpMux := http.NewServeMux()
+	httpMux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		lastPath = r.URL.Path
+	})
+
+	doubleSlashFixHandler := &DoubleSlashFixHandler{httpMux}
+
+	// Slash deduplication
+	{
+		lastPath = ""
+
+		req := httptest.NewRequest(
+			http.MethodGet, "http://example.com//v1/charges", nil)
+		w := httptest.NewRecorder()
+
+		doubleSlashFixHandler.ServeHTTP(w, req)
+
+		assert.Equal(t, http.StatusOK, w.Code)
+		assert.Equal(t, "/v1/charges", lastPath)
+	}
+
+	// Requests without duplicated slashes work normally
+	{
+		lastPath = ""
+
+		req := httptest.NewRequest(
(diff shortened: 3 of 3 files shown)


server.go
```diff
@@ -21,6 +21,25 @@ import (
...
+// The reason we deduplicate is that in some API libraries occasionally
+// generate paths with double slashes, and the real Stripe API responds to
+// these requests normally, so stripe-mock emulates that behavior.
+type DoubleSlashFixHandler struct {
+	Mux http.Handler
+}
+
+// ServeHTTP serves an HTTP request.
+func (h *DoubleSlashFixHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	r.URL.Path = strings.Replace(r.URL.Path, "//", "/", -1)
```
ob-stripe: I don't feel super strongly about it, but maybe we should collapse any number of consecutive slashes into a single one? Something like this:

```go
re := regexp.MustCompile(`/{2,}`)
r.URL.Path = re.ReplaceAllString(r.URL.Path, "/")
```

brandur-stripe: Thanks OB.

So this is kind of crazy, but while the API will take doubled-slashes, it actually won't take tripled slashes. e.g.

```
$ curl 'https://api.stripe.com//v1/customers/cus_FU5uUUsu4AU3k1' -u sk_test_xxx:
{
  "id": "cus_FU5uUUsu4AU3k1",
  "object": "customer",
  "account_balance": 0,
...

$ curl 'https://api.stripe.com///v1/customers/cus_FU5uUUsu4AU3k1' -u sk_test_xxx: -i
HTTP/2 404
server: nginx
date: Mon, 22 Jul 2019 22:59:24 GMT
content-length: 0
strict-transport-security: max-age=31556926; includeSubDomains; preload
```

So I figure we should just stick to a single deduplication pass to match that behavior.

Another pretty good reason is that I think the most common cause of a duplication is by appending two string variables together where you forget that you have both a trailing slash on the first as well as a leading slash on the other:

```
s1 = "https://api.stripe.com/"
s2 = "/v1/charges"
s1 + s2
```

That's pretty easy to do by accidentally, but getting a third slash in there is a little harder.

ob-stripe: Haha I didn't think to test it and just assumed whatever normalization we did would work for any number of slashes. 👍 to sticking to the API's behavior.

PR #176, opened 2019-08-14

remi-stripe:
Latest spec with latest API version

r? @ob-stripe 
cc @stripe/api-libraries


remi-stripe:
self-approving after chatting with ob!

PR #185, opened 2019-11-05

remi-stripe:
Releasing stripe-mock for the latest API version `2019-11-05`
cc @stripe/api-libraries


remi-stripe:
self-approving and releasing

PR #189, opened 2019-11-07
Use GoReleaser in Travis

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This PR enables the use of GoReleaser directly from Travis when a new tag is pushed. It also changes a few things in the GoReleaser configuration:
- Windows binaries are now compressed as `.zip` instead of `.tar.gz`
- `.deb` and `.rpm` packages are created
- the Docker image is created directly by GoReleaser instead of an external script

A few notes:
- the environment variables are now set in Travis' UI instead of the `.travis.yml` file
- the Github token used to create the release on Github belongs to my account (meaning all releases created from Travis will appear from my account in https://github.com/stripe/stripe-mock/releases). I'll reach out to dev-prod to see if we can get a token from a generic Stripe account like `stripe-ci` instead
- I tested this locally using `goreleaser --snapshot --skip-publish --rm-dist` and verified that the GoReleaser configuration works and the Docker image is built. The actual publish process is ~impossible to test locally though


.travis.yml
```diff
@@ -17,41 +14,15 @@ script:
   - go generate
   - make
 
-  # Build and test the Docker image. It's pushed to Docker Hub as part of the
-  # deploy phase.
-  - docker build -t "$DOCKER_REPO:${TRAVIS_COMMIT:0:8}" .
-  - docker run -d --name stripe-mock-container -p 12111:12111 "$DOCKER_REPO:${TRAVIS_COMMIT:0:8}"
-  - docker ps -a
```
brandur-stripe: Hm, just a note that there was some value to this — if you don't guarantee the integrity of the image here, then the only other way you find out something went wrong is finding out when a downstream build fails.

Dockerfile
```diff
@@ -1,30 +1,8 @@
...
-#
-# Use a tiny base image (alpine) and copy in the release target. This produces
-# a very small output image for deployment.
-#
 
 FROM alpine:latest
 RUN apk --no-cache add ca-certificates
-WORKDIR /
-COPY --from=builder /go/src/github.com/stripe/stripe-mock/stripe-mock .
+COPY stripe-mock /
```
brandur-stripe: Okay fine, I guess that seems easier, haha.

ob-stripe:
Going to merge this and attempt a release, and revert if something goes wrong.

PR #200, opened 2019-11-11
Allow `rk_` keys in addition to `sk_`

### LINKED ISSUE #199: Allow API Key to use restricted key
Currently stripe-mock returns an error when using a restricted key (rk_) vs using the standard secret key (sk_).  A very simple change would resolve this and allow for a developer to use either.  It is much more common (on my project anyway) for developers to be issued restricted keys so this would greatly improve the dev workflow.

Below is the code change needed (I tried to submit a PR but was unable):
server.go file - line 781-83
Current structure:
`if keyParts[0] != "sk" {
		return false
	}`
Updated fix:
`if keyParts[0] != "sk" || keyParts[0] != "rk" {
		return false
	}`

brandur-stripe:
Allow requests to be made to stripe-mock with the restricted access key
prefix (`rk_`) as well as the standard secret key prefix (`sk_`).

Fixes #199.

r? @ob-stripe
cc @stripe/api-libraries

### DIFF
--- a/server.go
+++ b/server.go
@@ -778,7 +778,7 @@ func validateAuth(auth string) bool {
 		return false
 	}
 
-	if keyParts[0] != "sk" {
+	if keyParts[0] != "rk" && keyParts[0] != "sk" {
 		return false
 	}
 


brandur-stripe:
Thanks!

PR #204, opened 2019-12-02
Add `-http-addr` and `-https-addr` options to listen on alternate IPs

### LINKED ISSUE #203: Let the user select the IP address to listen on
Currently stripe-mock listens on all available addresses which is not ideal in many context. Providing a CLI option would solve the issue and defaulting to localhost would also be great.

brandur-stripe:
Adds two new command line options that allow users to specify which IPs
they'd like stripe-mock to listen on.

Currently, only `-http-port` and `-https-port` are supported, which pass
an empty string to Go's `Listen` function, which causes stripe-mock to
listen on every available unicast and anycast address on the system [1].
This is fine for most users, but some want a little more control.

The glut of options available for binding to addresses is starting to
get a little out of control, so I've added a comment to the effect that
we're keeping existing options for compatibility, but that we should
consider deprecating `-http-port`, `-https-port`, `-port`, and `-unix`
to simplify the equation somewhat. Maybe at the same time we should
switch to Unix style flags too. (And maybe this will happen if/when
stripe-mock gets merged into the Stripe CLI.)

I've also made a few minor tweaks to the help strings for various
options to try and normalize conventions somewhat.

r? @ob-stripe Sorry, hope you don't mind me assigning this to you.
cc @stripe/api-libraries

Fixes #203.

---

[1] https://godoc.org/net#Listen

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,15 +36,28 @@ func main() {
 		httpsPortDefault: defaultPortHTTPS,
 	}
 
+	// As you can probably tell, there are just too many HTTP/HTTPS binding
+	// options, which is a result of me not thinking through the original
+	// interface well enough.
+	//
+	// I've left them all in place for now for backwards compatibility, but we
+	// should probably deprecate `-http-port`, `-https-port`, `-port`, and
+	// `-unix` in favor of the remaining more expressive and more versatile
+	// alternatives.
+	//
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Thanks OB!

PR #207, opened 2020-01-13
Adding convenience commands to Makefile for building and running docker image locally

joshuajlai:
Adding 2 commands to the Makefile:
- docker_build: runs the appropriate `docker build` command and tags the resulting image with the current commit hash as well as updating the `latest` tag
- docker_run: runs the appropriate `docker run` command with the default ports configured

If this PR is unnecessary, please feel free to close it. =)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+GIT_COMMITSHA = $(shell git rev-parse HEAD)
+IMAGE_NAME = "stripemock/stripe-mock"
+
 all: test vet lint check-gofmt
 
 check-gofmt:
@@ -14,14 +17,10 @@ test:
 vet:
 	go vet ./...
 
-# Docker commands
-.PHONY: docker_build docket_run
-
-GIT_COMMITSHA = $(shell git rev-parse HEAD)
-IMAGE_NAME = "stripemock/stripe-mock"
-
-docker_build:
+docker-build:
 	docker build -t "$(IMAGE_NAME):latest" -t "$(IMAGE_NAME):$(GIT_COMMITSHA)" .
+.PHONY: docker-build 
 
-docker_run:
+docker-run:
 	docker run --rm -it -p 12111-12112:12111-12112 "$(IMAGE_NAME):latest"
+.PHONY: docket-run
--- a/Makefile
+++ b/Makefile
@@ -13,3 +13,15 @@ test:
 
 vet:
 	go vet ./...
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -13,3 +13,15 @@ test:
 
 vet:
 	go vet ./...
+
+# Docker commands
+.PHONY: docker_build docket_run
```
brandur-stripe: For cleanliness, can we put the phony targets under each target? e.g.:

```
docker_build:
    docker build -t "$(IMAGE_NAME):latest" -t "$(IMAGE_NAME):$(GIT_COMMITSHA)" .
.PHONY: docker_builder
```

All the existing targets in here should have `.PHONY` as well (I'll add those later), and this will be cleaner for when it's a widely used convention.

Makefile
```diff
@@ -13,3 +13,15 @@ test:
 
 vet:
 	go vet ./...
+
+# Docker commands
```
brandur-stripe: Let's drop this comment — it doesn't really add any additional information given that the targets have a `docker*` prefix already.

Makefile
```diff
@@ -13,3 +13,15 @@ test:
...
+.PHONY: docker_build docket_run
+
+GIT_COMMITSHA = $(shell git rev-parse HEAD)
+IMAGE_NAME = "stripemock/stripe-mock"
+
+docker_build:
```
brandur-stripe: Can we rename these to use hyphens instead of underscores? We already have a hyphens with `check-gofmt` above, and I don't want to mix the two conventions.

Makefile
```diff
@@ -13,3 +16,11 @@ test:
...
 vet:
 	go vet ./...
+
+docker-build:
+	docker build -t "$(IMAGE_NAME):latest" -t "$(IMAGE_NAME):$(GIT_COMMITSHA)" .
+.PHONY: docker-build 
+
+docker-run:
+	docker run --rm -it -p 12111-12112:12111-12112 "$(IMAGE_NAME):latest"
+.PHONY: docket-run
```
brandur-stripe: One little misspelling here: "docket" should be "docker".

joshuajlai:
@brandonl-stripe Thanks for reviewing and catching that typo. It should be fixed now. Please let me know if I should add/change anything else. =)

brandur-stripe:
@joshuajlai You got it!

LGTM.

PR #211, opened 2020-02-05
Rebuild data replacement strategy to use input from OpenAPI schema

### LINKED ISSUE #210: default_tax_rates in subscription endpoint has the wrong type
Hi,

When testing the Java SDK against the mock server a call to  create a subscription (`POST http://mockserver/v1/subscriptions`) and setting the `default_tax_rates` parameter to a tax rate id (e.g. txr_123), the mockserver returns an subscription object containing a list of strings as `default_tax_rates` instead of an list of tax_rates (https://stripe.com/docs/api/subscriptions/object?lang=curl#subscription_object-default_tax_rates). 

This obviously leads to a JSON parse exception in the SDK.

Greetings
Georg

brandur-stripe:
Previously, the data replacer worked by taking input request and output 
response data, and replacing values if the types matched up. This generally
worked, but would result in some inappropriate replacements for more
complex data types like arrays, as seen in [1].

In this patch we upgrade the replacement strategy so that we perform 
replacements if the type of incoming request data matches what we expected
in the response based on the OpenAPI schema. This allows us, for example,
to not only check that a type is an array, but also what sort of elements
that array is supposed to contain, even if the value from our fixtures is
an empty array (which is often the case).

Fixes #210.

r? @ob-stripe
cc @stripe/api-libraries

---

[1] https://github.com/stripe/stripe-mock/issues/210

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -141,7 +141,11 @@ func (g *DataGenerator) Generate(params *GenerateParams) (interface{}, error) {
 	// simulate a more realistic create or update operation.
 	if params.RequestMethod == http.MethodPost {
 		if mapData, ok := data.(map[string]interface{}); ok {
-			mapData = datareplacer.ReplaceData(params.RequestData, mapData)
+			replacer := datareplacer.DataReplacer{
+				Definitions: g.definitions,
+				Schema:      params.Schema,
+			}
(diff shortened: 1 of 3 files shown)


brandur-stripe:
Thanks OB!

PR #257, opened 2021-07-08
Include /accept in URL paths which have a primary ID.

dcr-stripe:
Updates the list of suffixes for API methods with primary IDs to include `/accept`. This is in preparation of Quotes, which will have `/v1/quotes/qt_<id>/accept`. This ensure stripe-mock extracts and returns the primary ID for these calls.


remi-stripe (approved):
One day, we shall fix the need to do this fix :p

PR #259, opened 2021-07-08

richardm-stripe:
r? @dcr-stripe 

Contains quote.finalize -> quote.finalize_quote method rename


dcr-stripe (approved):
LGTM!

PR #261, opened 2021-07-20
Add support for search_result objects.

dcr-stripe:
r? @richardm-stripe 

Adds support for generating a `search_result` object with a single data element, rather than generating an empty result. This is similar to `list` behavior. As `search_result` has a slightly different schema, opted for duplicating the logic but happy to merge it in together.

This will need the updated OpenAPI spec before we can merge.

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -451,7 +451,7 @@ func (g *DataGenerator) maybeDereference(schema *spec.Schema, context string) (*
 	return schema, context, nil
 }
 
-func (g *DataGenerator) generateUrlForListableResource(schema *spec.Schema, params *GenerateParams) string {
+func (g *DataGenerator) generateURLForListableResource(schema *spec.Schema, params *GenerateParams) string {
 	var val string
 	if strings.HasPrefix(schema.Pattern, "^") {
 		// Many listable resources have a URL pattern of the form "^/v1/whatevers";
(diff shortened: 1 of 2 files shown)


richardm-stripe (approved):
LGTM!

generator.go
```diff
@@ -434,6 +451,22 @@ func (g *DataGenerator) maybeDereference(schema *spec.Schema, context string) (*
 	return schema, context, nil
 }
 
+func (g *DataGenerator) generateUrlForListableResource(schema *spec.Schema, params *GenerateParams) string {
```
richardm-stripe, line 454:
```suggestion
func (g *DataGenerator) generateURLForListableResource(schema *spec.Schema, params *GenerateParams) string {
```

generator.go
```diff
@@ -470,17 +503,7 @@ func (g *DataGenerator) generateListResource(params *GenerateParams) (interface{
...
-				// we cut off the "^" to leave the URL.
-				val = subSchema.Pattern[1:]
-			} else if params.example != nil {
-				// If an example was provided, we can assume it has the correct format
-				example := params.example.value.(map[string]interface{})
-				val = example["url"].(string)
-			} else {
-				val = params.RequestPath
-			}
+			val = g.generateUrlForListableResource(subSchema, params)
```
richardm-stripe, line 506:
```suggestion
			val = g.generateURLForListableResource(subSchema, params)
```

richardm-stripe (approved):

Issue #206, opened 2019-12-04
Refactor to make stripe-mock useable as a package


shoenseiwaso:
It would be great to be able to embed `stripe-mock` as a package in Go. For example, something like:
```go
package abc

import (
   "testing"
   "stripe/stripe-mock"
)

func TestMyApp(t *testing.T) {
   mock := stripe.RunMock()
   defer mock.Stop()

   // perform tests against stripe-mock here
}
```

In order to make this work, probably two changes are required:
- Move the bulk of the functionality out of `package main`. This could be accomplished either by moving most of the functionality in `github.com/stripe/stripe-mock` to another package in a sub-folder, or moving `package main` into a sub-folder (e.g., `github.com/stripe/stripe-mock/stripe-mock`). The former is probably preferred because no change would be required to anyone who has `go get -u github.com/stripe/stripe-mock` in their tooling.
- Go modularize `stripe-mock` (i.e., `go mod init` => `go.mod`, `go.sum`).

richardm-stripe:
This was shipped in #265

PR #238, opened 2020-08-31
Refactor project to use go modules and move the server inside of a new package.

Splizard:
The minumum work to fix #206

Now the server can be embedded inside of a Go project and integrated
with `go test`

Go generating the project directory bundles the certificates for main and the openapi data is now placed in the spec package which also houses the test data.

The project has been updated to use Go modules instead of the legacy Vendoring and I have updated the necessary tooling.

Some ideas that popped up while I wrote this (to cleanup the project):
- Move the certificate bundle and listener into the server package.
- Merge the server and spec packages together, so that testing functions don't need to be exported (stripetest package?).
- Have this repo or the stripe-go repo recognise the other with a simple way to enable the mock server for testing (ie in the way that #206 suggests)

PR #265, opened 2021-08-10
Merge changes to move server into its own package

akalinin-stripe:
## Notify
r? @richardm-stripe 
cc @dcr-stripe @stripe/api-libraries

## Summary
Based on the changes from https://github.com/stripe/stripe-mock/pull/238 (thank you @Splizard) with conflicts resolved against the master branch.

The original intent was to fix issue #206 and allow running embedded version of the server.
Detailed changes:
- Move server code from main package into /server directory
- Split embedded data generation to embed spec in server directory and certificates only in main directory
- Refactor out global state `verbose` variable and include it into server, so it does not become part of the API (not sure what to do with `Version` though, usually it is OK to leave it as it is set by ldflags).

Finally, it looks like it's a lot of changes, but mostly it's shuffling code around to a different package.

## Testing
- Unit tests updated
- Ran tests with the current stripe-go repo


akalinin-stripe:
False positive in golint is catching auto-generated file comment:
`sh
go list ./... | xargs -I{} -n1 sh -c 'golint -set_exit_status {} || exit 255'
/home/travis/gopath/src/github.com/stripe/stripe-mock/server/bindata.go:1:1: package comment should be of the form "Package server ..."
`

Makefile
```diff
@@ -9,10 +9,8 @@ build:
 check-gofmt:
 	scripts/check_gofmt.sh
 
-# This command is overcomplicated because Golint's `./...` doesn't filter
-# `vendor/` (unlike every other Go command).
 lint:
-	go list ./... | xargs -I{} -n1 sh -c 'golint -set_exit_status {} || exit 255'
+	golint
```
akalinin-stripe, line 13: golint is deprecated, and there's not much we can do to customize it's behavior. We have another task to replace it with a supported linter.

richardm-stripe: https://github.com/stripe/stripe-go/pull/1288 is what we did for stripe-go

akalinin-stripe: Updated to be the same in this repo.

main.go
```diff
@@ -1,21 +1,18 @@
-//go:generate go-bindata -mode 444 -modtime 1 cert/cert.pem cert/key.pem openapi/openapi/fixtures3.json openapi/openapi/spec3.json
+//go:generate go-bindata -mode 0444 -modtime 1 cert/cert.pem cert/key.pem
```
akalinin-stripe, line 1: Caught by staticcheck - permissions need to be in octal format.

main.go
```diff
@@ -1,21 +1,18 @@
...
+//go:generate go-bindata -o server/bindata.go -pkg server -mode 0444 -modtime 1 openapi/openapi/fixtures3.json openapi/openapi/spec3.json
```
richardm-stripe, line 2: does this mean we should delete `bindata.go` at the root? `server/bindata.go` is the new path?

server/server.go
```diff
@@ -133,13 +140,90 @@ type ResponseError struct {
...
 
+// LoadFixtures load fixtures from a JSON file
+//
+// If path is empty, fixtures are loaded from internal embedded assets.
+func LoadFixtures(fixturesPath string) (*spec.Fixtures, error) {
+	var data []byte
+	var err error
+
+	if fixturesPath == "" {
+		// And do the same for fixtures
```
richardm-stripe, line 151:
```suggestion
```

server/server.go
```diff
@@ -133,13 +140,90 @@ type ResponseError struct {
...
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("error loading fixtures: %v", err)
+	}
+
+	var fixtures spec.Fixtures
+	err = json.Unmarshal(data, &fixtures)
+	if err != nil {
+		return nil, fmt.Errorf("error decoding spec: %v", err)
```
richardm-stripe, line 168:
```suggestion
		return nil, fmt.Errorf("error decoding fixtures: %v", err)
```

richardm-stripe (changes requested):
This all looks good! The only thing is that I think we should remove `bindata.go` as the path has moved.

richardm-stripe (approved):
Clarified IRL! I understand the bindata commands now! Good to merge!

main.go
akalinin-stripe: Saving the comment for history here... The reason there are 2 bindata files is one of them is for fixtures/spec the other one is for the certs:
- server dir - includes fixtures3.json and spec3.json
- main dir - includes cert.pem and key.pem

PR #278, opened 2021-11-17

dcr-stripe:
r? @richardm-stripe 

Updates to account for `automatic_payment_methods`.


dcr-stripe:
Holding off on release.

PR #288, opened 2022-03-25
Use embed instead of bindata

pakrym-stripe:
GO has a native way of embedding files. Avoids having to run extra command to update specs.


akalinin-stripe (approved):
Looks like a good change.

pakrym-stripe:
I also checked the resulting file size, it went down by 400KB.

pakrym-stripe:
I'll update https://confluence.corp.stripe.com/display/DEVX/Releasing+stripe-mock when this goes in.

dcr-stripe (commented):
This looks good - thanks Pavel! Just to double check - how does this work when making a binary? Will `go:embed` bundle the file(s) into the binary like `go bindata` was doing?

pakrym-stripe:
> Will go:embed bundle the file(s) into the binary like go bindata was doing?

Yes, `go:embed` is a native replacement for `go bindata` was doing.

dcr-stripe (approved):
LGTM! Thanks Pavel - great stuff.

Ran a local test where I pulled branch, built using `make`, copied over the executable to an isolated folder, and all worked well.

PR #293, opened 2022-03-29

pakrym-stripe:
r? @dcr-stripe


dcr-stripe (changes requested):
There should be no .DS_STORE file. Let's remove that + add to gitignore.

pakrym-stripe:
PTAL @dcr-stripe ?

PR #295, opened 2022-04-14

richardm-stripe:


richardm-stripe:
Tests pass locally. Test failures in CI appear to be unrelated.

Issue #312, opened 2022-05-26
Response content-type should be application/json rather than text/plain


BlueHatbRit:
Hi, thanks for providing this mock project it's mega useful!

I found that when making an api request to the real stripe api, it responds with `application/json` as the respone content type, stripe-mock seems to respond with `text/plain` for everything which I don't think is correct.

Live API example:

```!sh
$ curl -i https://api.stripe.com
HTTP/2 404
server: nginx
date: Thu, 26 May 2022 09:15:48 GMT
content-type: application/json
content-length: 361
access-control-allow-credentials: true
access-control-allow-methods: GET, POST, HEAD, OPTIONS, DELETE
access-control-allow-origin: *
access-control-expose-headers: Request-Id, Stripe-Manage-Version, X-Stripe-External-Auth-Required, X-Stripe-Privileged-Session-Required
access-control-max-age: 300
x-content-type-options: nosniff
strict-transport-security: max-age=31556926; includeSubDomains; preload

{
  "error": {
    "message": "Unrecognized request URL (GET: /). If you are trying to list objects, remove the trailing slash. If you are trying to retrieve an object, make sure you passed a valid (non-empty) identifier in your code. Please see https://stripe.com/docs or we can help at https://support.stripe.com/.",
    "type": "invalid_request_error"
  }
}
```

Mock example:

```
$ curl -i http://localhost:12111
HTTP/1.1 401 Unauthorized
Stripe-Mock-Version: 0.130.0
Date: Thu, 26 May 2022 09:17:56 GMT
Content-Length: 255
Content-Type: text/plain; charset=utf-8

{
  "error": {
    "message": "Please authenticate by specifying an `Authorization` header with any valid looking testmode secret API key. For example, `Authorization: Bearer sk_test_123`. Authorization was ''.",
    "type": "invalid_request_error"
  }
}
```

While the responses do differ slightly due to the mocks implementation, this seems to happen for every API call to stripe-mock. When using some request libraries, they'll decode the response based on the content-type header which what made me stumble across this. It would be great if the mock could return the proper content type header, I _think_ in every single case this should be `application/json`?

dcr-stripe:
Thanks for the report @BlueHatbRit ! You're right that this is inconsistent with our main API. I've sent out https://github.com/stripe/stripe-mock/issues/312 to address this.

BlueHatbRit:
Thanks for the quick feedback and turn around on this @dcr-stripe it's much appreciated it!

PR #313, opened 2022-05-26
fix: Set Content-Type on responses.

dcr-stripe:
## Summary

Sets an appropriate `Content-Type` response header, mirroring our production API. Currently stripe-mock always returns JSON or PDF (it doesn't support the Files API).

## Motivation

Fixes https://github.com/stripe/stripe-mock/issues/312


pakrym-stripe (approved):

BlueHatbRit:
Thanks for this :)

PR #318, opened 2022-06-06
Attempt to use the standard github token with GoReleaser

pakrym-stripe:
CLI seems to be doing it: https://github.com/stripe/stripe-cli/blob/master/.github/workflows/release.yml#L73


dcr-stripe (approved):
LGTM!

PR #320, opened 2022-06-06
Add a workflow to auto-increment the tag

pakrym-stripe:
So stripe-mock can be released hands-free.

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -21,7 +21,7 @@ jobs:
         id: tag
         run: |
           last_tag="$(git describe --tags --abbrev=0)"
-          # Parse the version into parts, increment the minor version and reset the path to 0
+          # Parse the version into parts, increment the minor version and reset the patch to 0
           # Example:
           # v1.2.3 is parsed into $1 = "v1.", $2 = "2", $3 = ".3"
           # and concatenated back into v1.3.0
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -2,6 +2,9 @@ name: Create Tag
 on:
   workflow_dispatch:
     inputs:
+      tag:
+        description: "Tag name, defaults to minor increment."
+        required: false
       sha:
         description: "Commit SHA to tag. Defaults to latest."
         required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/tag.yml
```diff
@@ -0,0 +1,37 @@
+name: Create Tag
```
dcr-stripe, line 1: Let's make it obvious from the name that this will actively trigger a release.

.github/workflows/tag.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Set the commit SHA
+        id: sha
+        run: echo "::set-output name=commitSha::$(if [ -z "$SHA" ]; then echo $GITHUB_SHA; else echo $SHA; fi)"
+        env:
+          SHA: ${{ github.event.inputs.sha }}
+      - name: Set the tag
+        id: tag
+        run: |
+          last_tag="$(git describe --tags --abbrev=0)"
+          new_tag="$(echo "$last_tag" | perl -pe 's/^(v\d+\.)(\d+)(\.\d+)$/$1.($2+1).($3)/e')"
```
dcr-stripe, line 24: `perl` :)

dcr-stripe, line 24: This will keep the existing patch version which we probably don't want if bumping up the minor, right?

.github/workflows/tag.yml
```diff
@@ -0,0 +1,37 @@
...
+        env:
+          TAG: ${{ github.event.inputs.tag }}
+      - name: Create tag
+        uses: actions/github-script@v5
+        with:
+          script: |
+            github.rest.git.createRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'refs/tags/${{ steps.tag.outputs.newTag }}',
```
dcr-stripe, line 35: If we're going to support passing in an arbitrary tag, will this fail if we're re-using a tag that already exists?

dcr-stripe (commented):
Very nice! Just some small questions and then LGTM! 🚢 

Thanks Pavel!

.github/workflows/tag.yml
pakrym-stripe: Maybe? :)

.github/workflows/tag.yml
pakrym-stripe: It's just too powerful...

.github/workflows/tag.yml
dcr-stripe: Cool yeah if not let's add a failure here to catch that case and avoid accidentally stomping over an existing release.

.github/workflows/tag.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Set the commit SHA
+        id: sha
+        run: echo "::set-output name=commitSha::$(if [ -z "$SHA" ]; then echo $GITHUB_SHA; else echo $SHA; fi)"
+        env:
+          SHA: ${{ github.event.inputs.sha }}
+      - name: Set the tag
+        id: tag
+        run: |
+          last_tag="$(git describe --tags --abbrev=0)"
+          new_tag="$(echo "$last_tag" | perl -pe 's/^(v\d+\.)(\d+)(\.\d+)$/$1.($2+1).".0"/e')"
```
dcr-stripe, line 24: Nit: Let's add a comment explaining this regex since it isn't immediately obvious.

dcr-stripe (approved):
LGTM, just one nit.

PR #323, opened 2022-06-07
Fetch history when cloning the repo for tagging

pakrym-stripe:
The default `fetch-depth: 1` doesn't bring down all the tags. We need existing tags to determine the new tag.


dcr-stripe (approved):
LGTM - good catch!

Issue #281, opened 2021-12-01
Build for arm64 and publish multi-arch docker image to dockerhub


austin-tk:
This is useful when developing on Apple silicon as docker needs arm64 images.

This seems easy to augment following the example here https://goreleaser.com/customization/docker_manifest/?h=docker+mani#example-config

anniel-stripe:
Resolved in https://github.com/stripe/stripe-mock/pull/395

PR #380, opened 2023-01-08
Add support for multi-platform Docker images

sborsje:
Updates the `goreleaser.yml` config to support multi-platform Docker images (`amd64` and `arm64`). This is particularly useful for CI/CD pipelines that run on ARM instances.

(should resolve https://github.com/stripe/stripe-mock/issues/281)


pakrym-stripe (changes requested):
Thank you for your contribution @sborsje! 

Considering the docker manifest is an [experimental feature](https://docs.docker.com/engine/reference/commandline/manifest/) we wouldn't want to have it be the main way of delivering `stripe-mock` docker images.

What do you think about a separate arm64 image build that customers will have to reference manually for now (`stripe/stripe-mock:latest-arm64`)? When docker manifest goes GA we can reference it from the manifest.

pakrym-stripe:
Closing as inactive. Feel free to reopen!

pakrym-stripe:
> but manifests themselves are well specified and widely adopted
It seems I misunderstood the "experimental" notice on the `docker manifest` page.

We'll take another look at this.

PR #395, opened 2023-03-08
Fix multi arch release

ewdurbin:
Noticed #392. In order to build the amd64 image it's necessary to install qemu and buildx.

Validated on my fork: https://github.com/ewdurbin/stripe-mock/actions/runs/4364861761/jobs/7632707111#step:6:102

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,6 +18,12 @@ jobs:
           DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
         run: |
           echo "${DOCKER_PASSWORD}" | docker login --username "${DOCKER_USERNAME}" --password-stdin
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: 'arm64'
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
       - name: Set up Go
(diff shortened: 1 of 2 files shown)


anniel-stripe (approved):
Thanks so much for helping out with this!

ewdurbin:
Darn, looks like someone needs to update policies to allow the new third-party actions:

> docker/setup-qemu-action@v2 and docker/setup-buildx-action@v2 are not allowed to be used in stripe/stripe-mock. Actions in this workflow must be: within a repository that belongs to your Enterprise account, created by GitHub, or matching the following: goreleaser/goreleaser-action@v2, tibdex/github-app-token@v1.5.2.

😢

ewdurbin:
The allowed list should be at https://github.com/stripe/stripe-mock/settings/actions under "Actions permissions"

anniel-stripe:
Thanks for calling that out, I added docker/setup-buildx-action@v2 to the allowed list. It seemed like the QEMU set up step wasn't necessary, so I removed that in https://github.com/stripe/stripe-mock/pull/396

The release succeeds now, thanks again! 🎉

ewdurbin:
Interesting that qemu wasn't necessary, I suppose its because all the compiling is happening before the docker build.

Regardless, stoked to have an all arm dev env now 💜

PR #400, opened 2023-03-17
Support beta fixtures

pakrym-stripe:
To test beta APIs.

Binary size goes up from 16 to 20MB.

### DIFF
--- a/spec/spec.go
+++ b/spec/spec.go
@@ -82,6 +82,8 @@ var supportedSchemaFields = []string{
 	// This is currently being used to store additional metadata for our SDKs. It's
 	// passed through our Spec and should be ignored
 	"x-stripeResource",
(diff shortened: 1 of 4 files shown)


spec/spec.go
```diff
@@ -82,6 +82,8 @@ var supportedSchemaFields = []string{
 	// This is currently being used to store additional metadata for our SDKs. It's
 	// passed through our Spec and should be ignored
 	"x-stripeResource",
+	"x-stripeOperations",
+	"x-stripeParam",
```
pakrym-stripe, line 86: I'm using the SDK beta spec instead of `spec3.yaml`-like spec. I think we need to move to SDK specs across the board but we'll start with beta.

richardm-stripe (approved):

PR #402, opened 2023-03-17
Add fix goreleaser config and add release notifications

pakrym-stripe:
Yaml is hard


anniel-stripe (approved):
yaml is hard indeed 💀

PR #424, opened 2023-06-03
Replace :id with valid regex

anniel-stripe:
Fixes CI failure in https://github.com/stripe/stripe-mock/pull/423

### DIFF
--- a/spec/validate.go
+++ b/spec/validate.go
@@ -1,7 +1,9 @@
 package spec
 
 import (
-	"github.com/lestrrat-go/jsschema"
+	"strings"
+
+	schema "github.com/lestrrat-go/jsschema"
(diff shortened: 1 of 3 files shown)


spec/validate.go
```diff
@@ -1,7 +1,9 @@
 package spec
 
 import (
-	"github.com/lestrrat-go/jsschema"
+	"strings"
+
+	schema "github.com/lestrrat-go/jsschema"
```
anniel-stripe, line 6:
```suggestion
	"github.com/lestrrat-go/jsschema"
```

pakrym-stripe:
Should we fix the definition instead of working around it?

richardm-stripe (changes requested):
I think we should pursue fixing the issue in the spec instead

PR #432, opened 2023-06-30
Separate dockerfile for goreleaser

richardm-stripe:
Turns out #430 isn't compatible with goreleaser, which seems obvious in retrospect. I found [goreleaser docs](https://goreleaser.com/errors/docker-build) that suggest
> If you still want your users to be able to docker build without an extra step, you can have a Dockerfile just for GoReleaser, for example, a goreleaser.dockefile.

so, this PR is that.

### DIFF
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -59,6 +59,7 @@ nfpms:
     - rpm
 
 dockers:
+  - dockerfile: 'goreleaser.dockerfile'
   - image_templates:
       - "stripe/stripe-mock:{{ .Tag }}-amd64"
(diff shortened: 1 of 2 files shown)


goreleaser.yml
```diff
@@ -59,6 +59,7 @@ nfpms:
     - rpm
 
 dockers:
+  - dockerfile: 'goreleaser.dockerfile'
```
anniel-stripe, line 62: style nit
```suggestion
  - dockerfile: goreleaser.dockerfile
```

anniel-stripe (approved):

PR #437, opened 2023-07-13
Update installation command to use "go install" instead of "go get"(deprecated) in README.md file

### LINKED ISSUE #436: Update README with 'go install' instructions
I would like to suggest an improvement to the installation instructions in the README of the Stripe-Mock repository. Currently, the README suggests installing the package using the command 'go get', but I believe it would be more consistent and convenient to use 'go install' instead.

Here are the reasons for my suggestion:
1. Uniformity: Many Go packages recommend using 'go install' for installation, which is the standard way of installing Go binaries.
2. Simplicity: 'go install' downloads and installs the package in a single step, without the need for additional commands.
3. Reduced clutter: 'go install' places the binary in the default location, which is typically in the user's $GOPATH/bin directory. This avoids adding additional dependencies to the project.

I propose updating the README with the following installation instructions:

```shell
go install github.com/stripe/stripe-mock
```

I believe this change would enhance the user experience and align the installation process with the Go community's conventions. Thank you for considering my suggestion.

priyanshujain:
Starting in Go 1.17, installing executables with go get is deprecated. go install should be used instead.

This closes #436

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ API does, and might differ in nuanced (and potentially dangerous) ways.
 If you have Go installed, you can install the basic binary with:
 
 ```sh
-go get -u github.com/stripe/stripe-mock
+go install github.com/stripe/stripe-mock@latest
 ```
 
 With no arguments, stripe-mock will listen with HTTP on its default port of


anniel-stripe:
Thank you for the contribution!

PR #652, opened 2024-02-07
Allow partially override fixtures

Thomascogez:
### Context
First of all a big thanks to the **Stripe** team for making this tools, it allow us to implement **Stripe** services within our integration test with ease and confidence
### Motivation
99% of static mock responses are fitting our needs but we are currently working in depth with **BalanceTransaction** and some of it's mock keys are empty (eg: `fee_details`) . Since some of our test cases need those keys, I first try to use the `-fixtures` param with a custom fixture file that include only a custom `balance_transaction` object, but if the provided fixture does not implement every other resources mock they will be undefined and we want to keep our mock responses updated with latest open api specs.
### Proposed changes
This pr implement a new behavior for fixture loading that allow to merge `embedded` fixtures with `custom` fixtures (where custom override embedded), this allow to selectively provide custom resource mock
### Testing
I build a local docker image of this branch and test it with or integration test suite.
Should i had some unit tests for this feature ?

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -144,30 +144,42 @@ type ResponseError struct {
 //
 // If path is empty, fixtures are loaded from internal embedded assets.
 func LoadFixtures(embeddedFixtures []byte, fixturesPath string) (*spec.Fixtures, error) {
-	var data []byte
 	var err error
+	var rawCustomFixture []byte
 
-	if fixturesPath == "" {
-		data = embeddedFixtures
-	} else {
+	var marshalledCustomFixtures *spec.Fixtures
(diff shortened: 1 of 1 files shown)


anniel-stripe:
Hi @Thomascogez, for use cases that require a more sophisticated response than what stripe-mock provides, we encourage users to define their own mock as described in [the "Future Plans" section of our README](https://github.com/stripe/stripe-mock?tab=readme-ov-file#future-plans).

Unfortunately we won't be accepting this PR as we don't intend to add more sophisticated testing features to the library at this time.

PR #743, opened 2024-04-05
[Test] Use updated fixtures3.json

richardm-stripe:
(don't merge, just for running CI)


ramya-stripe:
@richardm-stripe Do we still need this to be open?

PR #1638, opened 2026-05-28
Fix vulnerable Go version in release workflow (fixes #1637)

### LINKED ISSUE #1637: stripe-mock builds against a vulnerable Go stdlib version (1.25.0)
I was running `stripe-mock` against Aqua scan at work.

It failed complaining about the `stripe-mock` binary being built against Go 1.25.0, which contains [numerous CVEs](https://www.cvedetails.com/version/2065866/Golang-GO-1.25.0.html), including the critical [CVE-2025-68121](https://www.cvedetails.com/cve/CVE-2025-68121/).

I've verified `stripe-mock` is built against Go 1.25.0 like so:
```bash
docker run -it --rm --entrypoint /bin/sh stripe/stripe-mock
apk update && apk add go
go version -m /bin/stripe-mock
```

which produces:
```
/bin/stripe-mock: go1.25.0
	path	github.com/stripe/stripe-mock
	mod	github.com/stripe/stripe-mock	v0.200.0+dirty
	dep	github.com/lestrrat-go/jspointer	v0.0.0-20181205001929-82fadba7561c
	dep	github.com/lestrrat-go/jsref	v0.0.0-20181205001954-1b590508f37d
	dep	github.com/lestrrat-go/jsschema	v0.0.0-20181205002244-5c81c58ffcc3
	dep	github.com/lestrrat-go/jsval	v0.0.0-20181205002323-20277e9befc0
	dep	github.com/lestrrat-go/pdebug	v0.0.0-20180220043849-39f9a71bcabe
	dep	github.com/lestrrat-go/structinfo	v0.0.0-20190212233437-acd51874663b
	dep	github.com/pkg/errors	v0.8.1-0.20170505043639-c605e284fe17
	build	-buildmode=exe
	build	-compiler=gc
	build	-ldflags="-s -w -X main.version=0.200.0 -X main.commit=497502e447781dc91b93d9d1b8f68d8d0f0db7c5 -X main.date=2026-05-26T19:31:13Z -X main.builtBy=goreleaser"
	build	CGO_ENABLED=0
	build	GOARCH=amd64
	build	GOOS=linux
	build	GOAMD64=v1
	build	vcs=git
	build	vcs.revision=497502e447781dc91b93d9d1b8f68d8d0f0db7c5
	build	vcs.time=2026-05-26T16:42:39Z
	build	vcs.modified=true
```

I see your Dockerfile [builds against golang:alpine](https://github.com/stripe/stripe-mock/blob/12d28f7/Dockerfile#L3), so maybe that's just cached somewhere, since the Go version in `golang:alpine` is currently 1.26.3:

```bash
% docker run -it --rm golang:alpine go version
go version go1.26.3 linux/amd64
```

edit: I guess that's not actually the correct Dockerfile, since it looks like [goreleaser](https://github.com/stripe/stripe-mock/blob/master/goreleaser.yml) builds the app. Maybe it's the goreleaser instance that needs to be set to a newer Go version

mbroshi-stripe:
### Why?

The release workflow pinned `go-version: 1.22.1` via the outdated `actions/setup-go@v2`. As reported in #1637, released binaries were being compiled with Go 1.25.0, which carries [CVE-2025-68121](https://www.cvedetails.com/cve/CVE-2025-68121/) and several other CVEs. The `ci.yml` workflow already used `go-version: stable` correctly — this brings the release workflow in line.

### What?

- **`release.yml`**: Switch `go-version` to `stable` so binaries always build against the current non-vulnerable Go toolchain. Also update stale actions: `checkout@v2→v4`, `setup-go@v2→v6`, `setup-buildx-action@v2→v3`, `goreleaser-action@v2→v6` (goreleaser v1.7→v2.16.0), and `--rm-dist→--clean` (goreleaser v2 rename).
- **`goreleaser.yml`**: Two required goreleaser v2 config migrations: `snapshot.name_template→version_template`, `brews[].tap→brews[].repository`.

## Changelog

- Fixed release workflow to build binaries against the current stable Go version instead of a pinned vulnerable version (resolves #1637).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ jobs:
       contents: read
     steps:
       - name: Code checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
           submodules: recursive
@@ -23,11 +23,11 @@ jobs:
         run: |
           echo "${DOCKER_PASSWORD}" | docker login --username "${DOCKER_USERNAME}" --password-stdin
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+        uses: docker/setup-buildx-action@v3
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v6
         with:
-          go-version: 1.22.1
+          go-version: stable
       # Fetch app token
       - uses: tibdex/github-app-token@v1.5.2
         id: gh-api-token
@@ -35,10 +35,10 @@ jobs:
           app_id: ${{ secrets.GH_APP_STRIPE_OPENAPI_APP_ID }}
           private_key: ${{ secrets.GH_APP_STRIPE_OPENAPI_PRIVATE_KEY }}
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v2
+        uses: goreleaser/goreleaser-action@v6
         with:
-          version: v1.7
-          args: release -f goreleaser.yml --rm-dist
+          version: ~v2.16.0
+          args: release -f goreleaser.yml --clean
         env:
           GITHUB_TOKEN: ${{ steps.gh-api-token.outputs.token }}
       - uses: stripe/openapi/actions/notify-release@master
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -20,7 +20,7 @@ builds:
       - arm64
 
 snapshot:
-  name_template: "{{ .Tag }}-next"
+  version_template: "{{ .Tag }}-next"
 
 archives:
   - format_overrides:
@@ -28,7 +28,7 @@ archives:
         format: zip
 
 brews:
-  - tap:
+  - repository:
       owner: stripe
       name: homebrew-stripe-mock
     homepage: "https://github.com/stripe/stripe-mock"
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -25,7 +25,7 @@ snapshot:
 archives:
   - format_overrides:
       - goos: windows
-        format: zip
+        formats: [zip]
 
 brews:
   - repository:


.github/workflows/release.yml
```diff
@@ -23,22 +23,22 @@ jobs:
...
           echo "${DOCKER_PASSWORD}" | docker login --username "${DOCKER_USERNAME}" --password-stdin
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+        uses: docker/setup-buildx-action@v3
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v6
         with:
-          go-version: 1.22.1
+          go-version: stable
```
jar-stripe, line 30: is there ever a world where a new version of go might deprecate or otherwise break on existing code?

jar-stripe (approved):
LGTM; the copilot feedback looks legit, and its worth a sanity check around using the stable version in CI

.github/workflows/release.yml
mbroshi-stripe: Not likely. Go's compatibility promise in https://go.dev/doc/go1compat is basically that they will always keep backwards compatibility, and in cases where they don't (e.g. bugs, security issues, or unspecified behavior), it's probably a feature that we would need to patch our code. I think there is a much bigger (and evidently real) risk of forgetting to update our Go version, and ending up with unpatched security issues.

.github/workflows/release.yml
```diff
@@ -23,22 +23,22 @@ jobs:
...
         with:
           app_id: ${{ secrets.GH_APP_STRIPE_OPENAPI_APP_ID }}
           private_key: ${{ secrets.GH_APP_STRIPE_OPENAPI_PRIVATE_KEY }}
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v2
+        uses: goreleaser/goreleaser-action@v6
         with:
-          version: v1.7
-          args: release -f goreleaser.yml --rm-dist
+          version: ~v2.16.0
```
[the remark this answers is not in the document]
mbroshi-stripe: Good catch. Updated `archives.format_overrides[].format` to a list (and kept to Go release v2).

PR #1643, opened 2026-08-14
Pin and modernize GitHub Actions references

jar-stripe:
### Why?

Resolves the 11 `unpinned-uses` findings zizmor reports for this repo by pinning GitHub action version to specific SHAs.  This PR also upgrades actions remove EOL versions and replace deprecated packages like `tibdex/github-app-token`.  Stacked on #1642.

### What?

* migrates use of deprecated `tibdex/github-app-token` to `actions/create-github-app-token@v3` to mint the publishing token
* adds `contents: write` on `stripe-mock` + `homebrew-stripe-mock` in `release.yml`, and `contents: write` on this repo alone in `tag.yml`
* upgrades `github-script` v5 to v9, `checkout` v2/v4 to v7, `setup-go` v6 to 7, `setup-buildx-action` v3 to v4, `goreleaser-action` v6 to 7.  Runtime and dependency bumps. No input removals that affect us.
* modifies tag inputs to read from `env`.  `${{ }}` no longer expands into the JavaScript body.
* awaits the `createRef` call in `tag.yml`.  the step now fails from the call site rather than through github-script's global unhandled-rejection handler.
* sets `persist-credentials: false` on all 3 checkouts.  the token no longer stays in `.git/config` after checkout.
* adds `.github/zizmor.yml` with a `ref-pin` policy scoped to the `stripe` namespace, so we can use `@master` with stripe owned actions.

**Verification.** 12 findings to zero findings (including finding fixed in #1642) reported by `zizmor 1.29.0` over `.github/workflows/`

### See Also
- #1642 — base branch; fixes this repo's `cache-poisoning` finding
- Reported by zizmor as `unpinned-uses`; this also clears `archived-uses`, `github-app`, `artipacked`, and `template-injection`: https://docs.zizmor.sh/audits/


.github/workflows/release.yml
```diff
@@ -12,40 +12,46 @@ jobs:
       contents: read
     steps:
       - name: Code checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
xavdid, line 15: interesting that we don't actually know what version we're using anymore (we just assume the comment is correct).

It's more secure, just an interesting side effect!

.github/workflows/release.yml
```diff
@@ -12,40 +12,46 @@ jobs:
...
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v6
+        uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
         with:
           version: ~v2.16.0
           args: release -f goreleaser.yml --clean
         env:
           GITHUB_TOKEN: ${{ steps.gh-api-token.outputs.token }}
-      - uses: stripe/openapi/actions/notify-release@master
+      - uses: stripe/openapi/actions/notify-release@325f3b157f7250f2a5d228b870d77bb63fc7e54c # v2407
```
xavdid: I mentioned this in the jira ticket, but IMO we shouldn't pin our own actions. (those actions should pin their dependencies internally). But I think we want fixes to get picked up automatically, especially for actions used across all the SDKs. This'll be doubly relevant when we start using the changelog stuff, which will be pretty GHA heavy.

xavdid (approved):
Looks good! Had a comment I feel fairly strongly about re: pinning our own actions, but am good otherwise!

.github/workflows/release.yml
jar-stripe: Yea.  Unfortunate IMHO.  It would be easier if github actions were in an "append only" style repo but tags can change and that's the risk :/  Keeping the intended version comment next to helps.

.github/workflows/release.yml
jar-stripe: That's fair.  I went ahead and reverted this and added the exception to the zizmor yml for Stripe owned actions.

